PR #23, opened 2026-05-22
docs(owasp): add per-entry anchors to LLM and Agentic tables

virtualsteve-exa:
## Summary

Adds a stable inline anchor (`#llm01` … `#llm10`, `#asi01` … `#asi10`) to each entry row in the OWASP LLM and Agentic tables in `docs/owasp.md`, so the GitHub Pages render is **deep-linkable to an individual entry**. The entries are table rows, and kramdown only emits anchor IDs for headings — so today `owasp.html` has only section anchors. The inline `<a id>` passes through kramdown without disturbing the compact table layout. (`RAISE.md` already has per-category heading anchors; this brings `owasp.md` to parity.)

**Why now:** this is a prerequisite for an upcoming 0.7.2 reporting change (on `dev`) that turns each finding's OWASP/RAISE tag capsule in the HTML report into a link to the exact entry in our own docs. Pages builds from `main`, so the anchors need to land here first — and landing them standalone lets us verify the deep-jumps actually resolve on the live site before the report feature depends on them. It also makes the live docs deep-linkable on its own merit.

Targets `main` directly (not `dev`) because it's a complete, standalone docs improvement we want published on Pages now, independent of the 0.7.2 reporting release.

## Testing

n/a — docs only. After merge + the Pages rebuild, I'll verify that `…/docs/owasp.html#llm01` and `#asi05` actually scroll to the right rows on the live site (the kramdown-inline-HTML-in-a-table-cell behavior is the thing being confirmed). Will report back on the thread.

## Notes for reviewers

- 20 rows changed, anchor-only; no gloss or wording edits.
- The dependent reporting change (tag capsules → links) is **not** in this PR; it lands on `dev` as part of 0.7.2.

### DIFF
--- a/docs/owasp.md
+++ b/docs/owasp.md
@@ -33,16 +33,16 @@ The risk landscape for any system that puts an LLM in the loop. Each finding Pra
 
 | ID | Risk | One-line gloss |
 |---|---|---|
-| **LLM01** | Prompt Injection | Untrusted input (direct or smuggled in via external content) overrides the model's instructions. |
-| **LLM02** | Sensitive Information Disclosure | The model leaks PII, secrets, or proprietary data through its outputs. |
-| **LLM03** | Supply Chain | Compromised or untrusted models, datasets, plugins, or dependencies enter the system. |
-| **LLM04** | Data and Model Poisoning | Training, fine-tuning, or RAG data is manipulated to bias or backdoor the model. |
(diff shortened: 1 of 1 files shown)


virtualsteve-exa:
Verified on live Pages after the rebuild (commit d7389ee): all 20 per-entry anchors render — `#llm01`–`#llm10` and `#asi01`–`#asi10` are present in `owasp.html`, and the inline `<a id>` passed through kramdown cleanly inside the table cells. Deep-links resolve. Thanks for the review.

Issue #32, opened 2026-05-26
RFE: Step 10 → `manifest_to_findings.py` — eliminate the canonical stall site by removing LLM composition from the JSON-emit phase


virtualsteve-exa:
## Context

The v0.7.3 release shipped a layered SKILL discipline aimed at the background-subagent no-progress watchdog stalls that dominated wallclock cost on the v0.7.3-prerelease Full Suite Run (~39% of attempts, see [`tests/runs/v0.7.3-prerelease/SUITE_RUN.md`](../blob/dev/tests/runs/v0.7.3-prerelease/SUITE_RUN.md)): a Step 9.9 full-prose manifest checkpoint, Step 10 mechanical-translation framing, skeleton-first Edit-append emission with one-line heartbeats. Subsequent Full Suite Runs (r4, r5-sonnet) showed those mitigations help but don't eliminate the stall pattern — they reduce the burst, but they don't remove the LLM composition phase that the watchdog is fundamentally aimed at.

The stall site is documented by the SKILL itself ([`SKILL.md:602`](../blob/dev/skills/behavior-verifier/SKILL.md#L602), [`SKILL.md:912`](../blob/dev/skills/behavior-verifier/SKILL.md#L912)): in a background subagent, ~600 s without a tool call kills the run, and the riskiest pause is the Step 10 JSON write because a complete multi-finding document is a long internal composition with nothing emitted to the tool-call stream until the `Write` (or each `Edit`) fires.

## Proposal

Add a deterministic `skills/behavior-verifier/manifest_to_findings.py` script that consumes the Step 9.9 draft manifest and emits canonical, schema-valid findings JSON. Replace the Step 10 LLM emission with a one-line invocation, mirroring how Step 11 already invokes `render.py`:

```bash
python3 "<SKILL_DIR>/manifest_to_findings.py" \
  --manifest ./reports/<agent-slug>-draft-<TIMESTAMP>.md \
  --out      ./reports/<agent-slug>-findings-<YYYY-MM-DD>.json
```

After the change:

- **Step 9.9** — LLM writes structured manifest (working artifact, parser-grade format).
- **Step 10** — Python script translates manifest → JSON (mechanical, deterministic, validates against `schema.py`).
- **Step 11** — Python script renders JSON → HTML+TXT (mechanical, deterministic, already exists as `render.py`).

## Why this is the right structural fix

Three facts the SKILL already establishes:

[...]

PR #30, opened 2026-05-25
Release 0.7.3: report v2, remit authoring, SKILL stall-prevention

virtualsteve-exa:
## Summary

The 0.7.3 release. Sixteen commits accumulated on `dev` since 0.7.2, grouped into four workstreams:

- **Report HTML redesign (#28)** — masthead, jump-nav, Executive Summary, collapsible findings; the visual overhaul completes the report-v2 work tracked in `feat/report-redesign`. Renderer + template + golden fixtures updated; all eleven `v0.7.0-sequential` regression baselines re-rendered byte-for-byte from their unchanged JSON.
- **Skill-assisted Worker Remit authoring** — the skill can now create / write / draft / build / author a remit from a repo or description, using `WORKER_REMIT_template.md` as required structure. SKILL frontmatter + Pre-flight section, with the template relocated to the skill directory after the first-attempt placement was wrong (issue #36, fix in `fix(skill): keep WORKER_REMIT_template at the repo root; redirect the pointers`).
- **SKILL stall-prevention** — four commits resolving the subagent watchdog stalls that gated v0.7.3-prerelease and r2 Full Suite Runs:
  1. `c8251c2` *stall-prevention: finding-prose discipline + emission discipline + troubleshooting* — tight per-finding prose caps + skeleton-first / Edit-append discipline + troubleshooting section
  2. `76176f3` *drop intermediate-render instruction* — `render every 2 findings` was incompatible with the renderer's count-consistency check
  3. `b733a45` *fix the fix — drop unrealizable skeleton-render check* — the skeleton `rules: []` failed the schema's min-1 constraint
  4. `88dd690` *enforce full-prose Step 9.9 manifest + mechanical Step 10 translation* — root-cause fix. Manifest must carry every prose value in final form; Step 10 is mechanical JSON-shape translation only.
- **Test plan codification + r3 validation** — `eb51d54` codifies the three-tier testing protocol (Smoke / Single-target / Full Suite Run) in `tests/README.md` and adds `tests/runs/` as the committed home for named pre-release runs; the v0.7.3-prerelease run is committed as the first entry. `2986846` adds the **post-SKILL-fix r3 Full Suite Run** at `tests/runs/v0.7.3-prerelease-r3/` — the authoritative validation evidence.

Plus six smaller docs / spec / README polish commits that are best viewed in the log.

## Testing

- **Smoke harness:** `python3 tests/render/test_render.py` — green locally; CI runs it on every push + PR across Python 3.9 / 3.12 / 3.13.
[...]

PR #34, opened 2026-05-27
0.7.4: deterministic Step 10 + emission discipline + version-source cleanup + rebaseline

virtualsteve-exa:
## Summary

Implements RFE #32 (deterministic Step 10) plus surrounding 0.7.4 work in three layers:

- **Deterministic Step 10.** `manifest_to_findings.py` mechanically converts the Step 9.9 parser-grade draft manifest to canonical findings JSON — no LLM in the loop. Replaces the historical LLM-composed JSON path that was the principal stall site on long scans. `schema.py` validates the emitted dict; a malformed manifest fails fast with the offending path.
- **Step 9.9 chunked-write emission discipline + Step 3 pacing guidance.** Manifest emission is now skeleton-first + Edit-append + heartbeats, eliminating the silent-compose burst that tripped the 600 s subagent no-progress watchdog at scale. Step 3 picks up a one-line pacing guard against long internal-only reasoning between large source reads.
- **Version source of truth cleanup.** `manifest_to_findings.py` reads `praxen_version` from `.claude-plugin/plugin.json` and `schema_version` from `schema.SCHEMA_VERSION`. SKILL no longer writes either; the manifest header drops both lines. Old manifests with the two fields parse cleanly (accepted-and-discarded) for backward compat. `build.sh` now sanity-checks that `PRAXEN_SPEC.md`, `plugin.json`, and `marketplace.json` agree on the version — future drift fails the build.

Version bumped to 0.7.4 across `plugin.json`, `marketplace.json`, `PRAXEN_SPEC.md`, and `README.md` (was stale at 0.7.2). CHANGELOG stub added.

Rebaselined as [`tests/baselines/v0.7.4-sequential/`](https://github.com/open-ai-security/praxen/tree/feat/manifest-to-findings/tests/baselines/v0.7.4-sequential): 11 cold runs, full BASELINE.md narrative, per-target bands in `tests/README.md` recalibrated against the new baseline. `v0.7.0-sequential/` retired (kept on disk for diff archaeology).

## Full Suite — c88e71c

11/11 success, **0 stalls**, median wall ~8 min (range 5–11 min). Previous v0.7.3-era runs were 12–18 min with ~39 % stall rate — roughly half the wall time, no stalls. Every target carries ≥ 2 Mediums (Medium-tier completeness check holds). Aider and OpenHands (historically flakiest) cleared without rescue. Per-target severity + RAISE table in `tests/baselines/v0.7.4-sequential/BASELINE.md`.

## Test plan

- [x] `python3 tests/render/test_manifest_to_findings.py` — 25/25 pass
- [x] `python3 tests/render/test_render.py` — 176/176 pass (includes 33 new baseline artifacts + verbatim-quote check against `tests/remits/<slug>.md`)
- [x] `bash build.sh` — produces `praxen-0.7.4.zip`; version-agreement guard active
- [x] Smoke-tested the new converter against an existing draft manifest (helperbot from suite output): emits canonical `praxen_version: 0.7.4` regardless of what the manifest header said
- [ ] **CI (Python 3.9 / 3.12 / 3.13)** — will run on this PR
- [ ] **Gemini review**
- [ ] Plugin-install smoke check pre-tag: `claude plugin marketplace add open-ai-security/praxen` + `install praxen@open-ai-security`

## Out of scope / follow-ups

- Per-target band widths in `tests/README.md` are centered on this single v0.7.4 run with the same ±2 counts / ±0.3–0.4 weighted shape used for v0.7.0. Future re-runs will refine the widths; this is the new starting point.
- Plugin-install smoke check pre-tag (per `feedback_test_plugin_install_before_release`).


virtualsteve-exa:
Thanks for the review. All four medium-priority comments addressed in c6f5e80.

Consolidated the four call sites behind a single `_is_null_sentinel(v)` helper — case-insensitive match for `""` / `"null"` / `"none"` — applied at:

- `related_findings` parse-time (where the existing empty-string handling lived)
- `finding_id` (rules), `owasp_llm`, `owasp_agentic`, `policy_rule_ids` post-parse in `_populate_derived`, alongside the existing defaulting logic

Locked the contract in `test_manifest_to_findings.py` Section 5b — mutate the helperbot fixture to write `none` for a rule's `finding_id` and `NONE` for a finding's `related_findings`, assert both normalize. Tests now 28/28 (was 25/25) + 176/176 render.

virtualsteve-star:
## Validation review — PR #34 (v0.7.4: deterministic Step 10)

**Context**: I ran the v0.7.3-dev validation earlier this week (PR #30) against two real-world targets — Google's ADK ambient-expense-agent and Lobot. That work uncovered the exact failure modes this PR is designed to eliminate. Reviewing from that firsthand perspective.

---

### What this PR fixes and why it matters

**The root cause** in our validation runs was that Step 10 was LLM-composed JSON. That made the final output the most variable and stall-prone step in the pipeline. We hit two concrete bugs:
- r1 Lobot: weighted_overall arithmetic error (1.85 vs 1.75) — caught by pre-render self-check, but required manual correction
- r2 Lobot: `stat_counts` mismatch on first render attempt (`verified=5` but `rules[]` had 4 entries) — caught by the new tally instruction, corrected on retry

Both bugs come from asking an LLM to maintain running counts while composing JSON. With `manifest_to_findings.py`, the converter owns that arithmetic. These errors become structurally impossible.

The stall-prevention angle is equally significant. The subagent watchdog (~600s) historically killed long scans. A deterministic converter is fast and predictable; an LLM composing JSON from scratch can stall anywhere. The manifest discipline + mechanical translation approach in this PR closes the gap that caused the historical ~39% stall rate.

---

### `manifest_to_findings.py` — code review

The implementation is solid. Specific observations:

**`_is_null_sentinel()`** — The Gemini review correctly flagged this gap. LLMs routinely write `"null"`, `"none"`, or `""` when they mean JSON `null`, and these arrive as Python strings after parsing. The helper correctly normalizes all three cases (plus case variations) to `None`. Applied at all the right call sites: `related_findings`, `finding_id`, `owasp_llm`, `owasp_agentic`, `policy_rule_ids`. Good fix.

**`_load_praxen_version()`** — Reads from `.claude-plugin/plugin.json` relative to `__file__`. The comment documents the assumption: `skills/` and `.claude-plugin/` share a common plugin root in both dev repo and installed deployments. That assumption holds for the current layout. The hard-fail behavior is correct — silently emitting a stale version string defeats the point of the version-source-of-truth cleanup.

One worth-noting edge case: if someone runs `manifest_to_findings.py` from a copy of `skills/` that's outside the plugin layout (e.g., a manual checkout into a different directory structure), `_load_praxen_version()` will raise `FileNotFoundError` with a path that references `../../.claude-plugin/plugin.json`. The error message could make the path assumption explicit — something like `"expected at {pp} (two levels above skills/behavior-verifier/)"`. Not a blocker, but would save debugging time for anyone who copies the script.

**`MANIFEST_FORMAT_VERSION = 1`** — Good to have a machine-readable contract version. The parser enforces it (line 465). Future manifest format changes will fail loudly rather than silently producing wrong output.

**Canonical key ordering** (lines 1100–1111) — Intentionally building the output dict in schema-declaration order rather than using `sort_keys`. This is the right choice: `sort_keys` would produce alphabetical order which diverges from schema declaration order, making diffs harder to read. The test at line 316 (key order assertion) verifies this.

**`stat_counts` and weighted_overall** — Both are computed from ground truth (rules array and category scores), not passed through from the manifest. This is the structural fix for the two bugs we hit. I confirmed by tracing through the code: neither field has a manifest field-type entry; they're built in `_finalize_remit_coverage()` and `_finalize_raise_posture()`.

**Error messages** — `ManifestError` messages include line numbers throughout. This is important: when the converter rejects a manifest, the worker needs to find the offending line quickly. Good discipline.

**Stdlib-only** — Correct for a tool that needs to work in fresh Claude Code sessions without pip installs. No new dependencies introduced.

---

### SKILL.md changes

**Step 9.9 parser-grade manifest spec** — This is the right architectural move. The manifest is now a contract between the worker and the converter, not just a freeform draft. The field-by-field prescription ensures the converter can parse mechanically. The key discipline here: the worker writes the manifest; the converter, not the worker, handles all arithmetic (counts, weighted scores). The SKILL change correctly separates these concerns.

**"Don't tier-compress" Medium guidance** — The XSS finding in our r1 run was a direct casualty of this: the HITL-bypass consequence was real but the compound impact wasn't named, so the finding landed as two High findings rather than one Critical. The Step 7 compound-escalation hedge in 44057a8 was a behavioral fix; this is the calibration fix that should prevent the pattern upstream. Directly connected to our validation experience.

**Step 3 pacing guard** — The trimmed prose is cleaner. The prescription (read KB files before scoring, no exception) is preserved. Good edit.

**Step 9.6 simplification** — Makes sense; the manifest discipline in 9.9 makes the earlier checkpoint lighter.

---

### Test coverage

Ran both suites locally on this branch:

```
python3 tests/render/test_manifest_to_findings.py
28 passed, 0 failed

python3 tests/render/test_render.py
176 passed, 0 failed
```

The 28 converter tests cover: happy-path round-trip against the helperbot golden fixture, determinism, schema validation, render integration, value coercion (null for nullable fields, sentinel normalization), 7 negative cases, and canonical key order. Good coverage for a new module.

One minor observation: the test file is a standalone script (`sys.exit()` at module level, no pytest dependency). This is intentional and documented in the header. However, if someone tries to run `pytest tests/render/test_manifest_to_findings.py` they'll get `INTERNALERROR: SystemExit: 0` from pytest importing the module at collection time. The header doc is clear about the correct invocation, but a pytest guard (`if __name__ == "__main__": sys.exit(...)`) would prevent the confusing error for anyone who tries the wrong runner. Minor quality-of-life issue, not a blocker.

---

### Connection to PR #30 open items

**stat_counts manual tally errors** — **Fully resolved**. The converter computes these from ground truth. No more manual arithmetic in Step 10.

**Remit rule extraction variance** (same prose remit → 30/13/12 rules across three runs) — **Still open**. This is a calibration/SKILL issue, not a Step 10 issue. The three approaches I flagged in PR #30 (numbered rules in template, granularity calibration in Step 2, document as expected variance) remain candidates. This PR doesn't address it and doesn't need to.

**`enp` vs `gap` for absent modules** — **Still open**. Same situation — not Step 10's domain.

---

### Overall assessment

The architecture is right: a deterministic, stdlib-only converter between the LLM's freeform work (the manifest) and the structured output (findings JSON) eliminates an entire class of correctness and stall problems. The implementation is clean, well-tested, and hard-fails on structural violations. The `_is_null_sentinel` fix from the Gemini review was the right call and is well-targeted.

The one path-assumption in `_load_praxen_version()` is worth a comment improvement but not a blocker.

Ready to merge from my perspective. The two remaining open items are SKILL-level concerns for a future PR.

_Tested on: feat/manifest-to-findings @ c6f5e80, Python 3.13, macOS_

virtualsteve-exa:
Thanks for the thorough validation review @virtualsteve-star — and great to hear the architecture lines up with what r1/r2 surfaced on ambient-expense-agent and Lobot.

Both QoL flags addressed in 42c44c4:

1. **`_load_praxen_version()` error message** — bare `FileNotFoundError` on the resolved path is now wrapped in a `RuntimeError` that names the expected location relative to `skills/behavior-verifier/` and tells the reader where the script is supposed to live. Same hard-fail semantics, clearer message for the "someone copied the script outside the plugin tree" case you flagged.

2. **`test_manifest_to_findings.py` pytest guard** — wrapped the final `sys.exit()` in `if __name__ == "__main__":`. As you noted, the file is a standalone script (no pytest dependency by design), but the wrong-runner case now falls through to pytest's "no tests collected" message instead of the `INTERNALERROR: SystemExit` confusion.

Both 28/28 converter + 176/176 render still green.

The two open items you noted from PR #30 (remit-rule-extraction variance, `enp` vs `gap` for absent modules) are SKILL-level calibration concerns and out of scope for this PR — agreed. Worth their own follow-up.

Issue #41, opened 2026-05-28
Skill gap: no named detection pattern for external API response → filesystem write (path traversal class)


virtualsteve-star:
## Summary

A Praxen scan of Hermes Agent v0.15.0 missed CVE-2026-7396, a path traversal vulnerability in the WeChat platform adapter (`gateway/platforms/weixin.py`). Post-scan cross-reference with a third-party security report (Repello AI) surfaced the CVE; manual code review confirmed it is unpatched. This issue documents the root cause of the miss and proposes three concrete improvements to the skill.

---

## The Vulnerability That Was Missed

**CVE-2026-7396** — path traversal in `gateway/platforms/weixin.py`.

The WeChat adapter receives `ilink_bot_id` from the iLink WeChat API response during the QR authentication flow and uses it unsanitized in filesystem path construction:

```python
# weixin.py:1149 — account_id sourced directly from external API response
account_id = str(status_resp.get("ilink_bot_id") or "")

# weixin.py:230-231 — used directly in path join, no validation
def _account_file(hermes_home: str, account_id: str) -> Path:
    return _account_dir(hermes_home) / f"{account_id}.json"
```

The same unsanitized value flows into two additional path construction sites:
- `ContextTokenStore._path()` at line 276: `self._root / f"{account_id}.context-tokens.json"`
- `_sync_buf_path()` at line 1023: `_account_dir(hermes_home) / f"{account_id}.sync.json"`

`atomic_json_write()` in `utils.py` calls `path.parent.mkdir(parents=True, exist_ok=True)` before writing, so a traversal value like `../../../.ssh/evil` causes Hermes to create intermediate directories and write outside `~/.hermes/weixin/accounts/`. The `.json` suffix appended at each call site limits direct overwrite of exact-named sensitive files but does not prevent writes to arbitrary user-writable paths.

The same `account_id` is also sourced from operator config/env (`WEIXIN_ACCOUNT_ID`) — that path is within the trust envelope — but the iLink API response path is not.

**Confirmed unpatched in v0.15.0** (cloned 2026-05-28). No `resolve()`, `is_relative_to()`, or character-filtering call exists in any of the three path construction sites.

The one-line class of fix:
```python
def _account_file(hermes_home: str, account_id: str) -> Path:
    base = _account_dir(hermes_home)
    path = (base / f"{account_id}.json").resolve()
    if not path.is_relative_to(base.resolve()):
[...]

PR #42, opened 2026-05-28
skill(pre-flight): multi-component remit guidance + flag-not-restrict rule

virtualsteve-star:
## Summary

Two improvements to the Pre-flight remit authoring procedure, motivated by a real session where an under-documented SSH management feature was incorrectly prohibited in the remit, producing a Critical finding that was a remit documentation error rather than a code vulnerability.

- **Step 5 — Flag under-documented capabilities rather than assuming restrictions.** When docs name a capability without scoping it, write the best inference tagged `[Inferred]` and surface it as an open question at delivery — rather than writing a MUST NOT clause based on an assumed scope.
- **Multi-component deployments paragraph** — guidance on when to combine vs. split remits, and how to structure a combined remit within the fixed template: scope note in Mission, sub-headings within existing sections, no new top-level sections.
- **`WORKER_REMIT_template.md` Mission placeholder** updated to surface the multi-component guidance at point of authoring.

## Related

Companion to issue #41 (external input → filesystem write detection pattern), which surfaced from the same scan session.

## Test plan

- [ ] Read the updated Pre-flight section and confirm steps 1–5 read as a coherent sequence
- [ ] Confirm the Multi-component paragraph sits naturally between step 5 and the "request that is only authoring" closing line
- [ ] Confirm the template Mission placeholder change doesn't break any existing template tests


skills/behavior-verifier/SKILL.md
```diff
@@ -103,6 +103,9 @@ If the operator's request is to **create, write, draft, build, or author a Worke
...
 4. **Deliver the remit** at the operator's preferred path (default: `WORKER_REMIT.md` in the current directory, or `WORKER_REMIT_<AGENT>.md` for multi-agent workspaces).
+5. **Flag under-documented capabilities rather than assuming restrictions.** If documentation names a capability without scoping it — *"supports SSH tunnel mode"*, *"executes shell commands"* — write your best inference of the scope, tag it `[Inferred]`, and flag it as an open question when you deliver the remit.
+
+**Multi-component deployments.** When covering more than one cooperating component in a single remit, use separate remits if the components can be independently deployed and audited; combine only when they are tightly coupled and only make sense together. If combining, the template structure must be preserved exactly — Phase 1 inventory walks known section headings, so invented top-level sections break the parse. Use a scope note in the Mission section to name each component and designate which is the primary RAISE subject, and sub-headings within existing sections (`### Component A` / `### Component B`) to separate per-component rules.
```
virtualsteve-exa: Phase 1 explicitly says "*adapt to whatever sections the remit uses*" (`SKILL.md:358`) — it's heuristic LLM guidance, not a parser. "Break the parse" overstates what actually happens: the practical effect of an invented section is that rules in it get *missed*, not that anything literally fails. Suggested softer wording below.

```suggestion
**Multi-component deployments.** When covering more than one cooperating component in a single remit, use separate remits if the components can be independently deployed and audited; combine only when they are tightly coupled and only make sense together. If combining, the template structure must be preserved exactly — Phase 1 inventory looks for the standard section headings (and adapts to whatever sections the remit uses), so rules placed in invented top-level sections risk being skipped. Use a scope note in the Mission section to name each component and designate which is the primary RAISE subject, and sub-headings within existing sections (`### Component A` / `### Component B`) to separate per-component rules.
```

virtualsteve-exa (commented):
## Review — PR #42

**Recommend merge with one minor wording suggestion** (inline on `SKILL.md:108`) and one optional follow-up.

### What's well-handled

- **Step 5 fills a real gap.** Existing Step 3 covers the *silent-section* case ("ask one targeted question rather than guess"); the new Step 5 covers the distinct *named-but-unscoped* case ("supports SSH tunnel mode" with no scope) — write the inference tagged `[Inferred]`, flag at delivery. The two patterns are semantically different and the new step preserves the distinction. Example phrases are well-chosen.
- **Multi-component guidance is consistent.** The combine-vs-split criterion ("independently deployed and audited" vs "tightly coupled and only make sense together") matches the implied intent of designating a single primary RAISE subject. Structural rule (sub-headings within existing sections; no new top-level sections) is the right call given how Phase 1 actually works. Surfacing the guidance in the template placeholder ensures the operator sees it at point of use.
- **No cross-doc drift.** Placeholder text isn't referenced elsewhere; no overlap with existing SKILL passages; `PRAXEN_SPEC.md` doesn't describe pre-flight authoring, so no spec sync needed.
- **Engine impact: none.** `tests/render/test_render.py` passes 176/0 on the branch. Schema, scoring, baselines all unaffected.

### Optional follow-up (not blocking)

`docs/writing-remits.md` doesn't yet reflect the multi-component guidance or the flag-don't-restrict rule. Its "Common mistakes" and "Required sections" headings would be natural landing spots — worth a separate doc-sync PR rather than expanding this one, same shape as the doc-sync follow-up for the OWASP grids.

_Tested on: feat/multi-component-remit-guidance, Python 3.13, macOS_

PR #49, opened 2026-05-29
chore: sync dev with main (catch up through 0.7.7)

virtualsteve-exa:
## Catch `dev` up to `main` (through 0.7.7)

`dev` had fallen **two releases behind `main`** — it was at `0.7.5`-era content and was missing the entire **0.7.6** and **0.7.7** releases (no `v0.7.7-sequential` baseline, `plugin.json` at `0.7.5`, CHANGELOG topping out at `[0.7.5]`). The 0.7.6/0.7.7 release PRs landed on `main` but were never back-merged into `dev`.

This is a **back-merge of `origin/main` into `dev`** to restore the intended `dev → main` flow. `main` is a strict content superset of `dev` (nothing on `dev` is lost — verified: zero files exist on `dev` that `main` lacks), so the merge result tree is **byte-identical to `main`** (`git diff origin/main HEAD` → 0 lines). Conflicts were the expected squash-divergence kind and were resolved in favour of `main` (the newer release line).

### After this merges
- `dev` content == `main` content (version `0.7.7`, `v0.7.7-sequential` baseline present, CHANGELOG `[0.7.7]`).
- The follow-up **`v0.7.7-claude48` rebaseline** PR can then target `dev` with a clean 2-commit diff.

### Validation
- `python3 tests/render/test_render.py` → **242 passed, 0 failed**.
- No skill/schema/engine change — this is purely a branch-sync.


virtualsteve-exa:
Thanks @gemini-code-assist. Two notes on the `owasp_coverage.py` `as_uri()` finding:

**1. Nothing to change in this PR.** #49 is a pure catch-up back-merge of `main` into `dev` — its tree is byte-identical to `main` (`git diff origin/main HEAD -- tests/baselines/owasp_coverage.py` → 0 lines). The flagged code shipped in 0.7.7; editing it here would re-diverge `dev` from `main`, defeating the sync. Any change to that file belongs on a normal feature branch.

**2. On the merits it's a false positive — no `AttributeError` path exists.** `info["report"]` is only ever a `Path` or `None`:
```python
"report": analysis_html[-1].resolve() if analysis_html else None,   # gather()
```
and the `None` case is excluded by the guard before the `try`/`except`:
```python
if info.get("report"):
    try:
        rel = Path(os.path.relpath(info["report"], out_dir)).as_posix()
    except ValueError:                 # only Windows cross-drive relpath
        rel = info["report"].as_uri()  # info["report"] is a Path here → valid
```
So inside the `except`, `info["report"]` is always a `Path` and `.as_uri()` is safe. There's no code path where it's a `str`.

If you'd still like the defensive `Path(info["report"]).as_uri()` wrap as belt-and-suspenders, I'm happy to fold it into a separate small chore (the upcoming `v0.7.7-claude48` rebaseline PR already touches this file) — but it's not required for correctness, and it doesn't belong in this back-merge.

virtualsteve-exa:
Logged the `as_uri()` hardening as #50 to clean up in a later chore batch — keeping this back-merge content-identical to `main`.

Issue #50, opened 2026-05-30
bug: owasp_coverage.py — harden as_uri() fallback against non-Path report values


virtualsteve-exa:
### What happened?

In `tests/baselines/owasp_coverage.py`, `target_cards()` has a fallback path that calls `.as_uri()` directly on `info["report"]`:

```python
if info.get("report"):
    try:
        rel = Path(os.path.relpath(info["report"], out_dir)).as_posix()
    except ValueError:                 # Windows cross-drive relpath
        rel = info["report"].as_uri()  # assumes info["report"] is a Path
```

`.as_uri()` is a `pathlib.Path` method — if `info["report"]` were ever a plain `str`, this would raise `AttributeError`. Surfaced by automated review on PR #49.

### Severity / current impact

**Latent, not currently triggerable.** `info["report"]` is only ever set to a `Path` (`analysis_html[-1].resolve()`) or `None`, and the `None` case is excluded by the `if info.get("report"):` guard — so today the `except` branch always has a `Path` and `.as_uri()` is safe. There is no input that reaches this line with a `str`. This is a **defensive-hardening / robustness** item, not an observed failure: a future refactor that let `report` hold a string (e.g. reading paths from JSON/config) would turn it into a real crash, on the rarely-exercised Windows cross-drive `except` branch.

### What did you expect to happen?

The fallback should be robust to `report` being a path-like string, e.g. wrap it: `rel = Path(info["report"]).as_uri()`.

### How to reproduce

Not reproducible with current inputs (see above). Would require `info["report"]` to be a `str` reaching the `except ValueError` branch (Windows, cross-drive `out_dir` vs report path).

### Praxen version

0.7.7 (code is unchanged on `main`; same in the pending `v0.7.7-claude48` rebaseline branch).

### Environment

`tests/baselines/owasp_coverage.py` — stdlib-only baseline OWASP-coverage report generator.

### Additional context

- Raised by `@gemini-code-assist` on PR #49: https://github.com/open-agent-ai-security/praxen/pull/49#issuecomment-4580768287
- Fix is a one-liner (`Path(info["report"]).as_uri()`). Deferred for a later cleanup batch rather than threaded into the dev-sync back-merge (#49) or the rebaseline PR. Good candidate for a chore-batch branch.

virtualsteve-exa:
[...]

Issue #48, opened 2026-05-29
RFE: RAISE scoring rigour — reduce undue per-category variance via a structured control-ledger + boundary decision rules


virtualsteve-exa:
## What is the problem you're trying to solve?

When the **same target is re-scanned with the same skill version and the same Worker Remit**, the *finding set* is highly reproducible but the **weighted RAISE score is not** — and the variance lives almost entirely in the per-category 0–5 scoring, specifically in how much credit *operative-but-imperfect controls* earn at the 0↔1 and 2↔3 boundaries.

This was measured directly. Re-running four targets 3× each on a fixed skill (v0.7.7), holding sources and remits constant and changing only the model (Opus 4.7 → 4.8):

| Target | run1 | run2 | run3 | mean | range |
|---|--:|--:|--:|--:|--:|
| openai-customer-service | 1.60 | 1.05 | 1.20 | 1.28 | **0.55** |
| aider | 2.00 | 2.15 | 2.15 | 2.10 | 0.15 |
| langchain-sql | 1.45 | 1.60 | 1.45 | 1.50 | 0.15 |
| sweep | 1.45 | 1.30 | 1.30 | 1.35 | 0.15 |

The openai-customer-service swing (0.55, well past the ~±0.3 we treat as expected) is the clearest case: **the finding set was identical across all three runs** (guardrails-shipped-but-unused, `on_seat_booking_handoff` fabricating a flight number via `random.randint()`, raw-model-arg seat mutation, audit log absent), yet the weighted score moved 0.55 because the judgment "how much does the SDK's default tracing / strict-schema tool-args credit the example agent?" landed differently each run. `tests/README.md` already documents this exact target as a "0.6↔1.8 swing."

The practical cost: a single full-suite run can catch several targets at simultaneous high (or low) draws and **mis-state where a target really sits relative to its band** — which directly undermines the release-gate comparison and any attempt to freeze a stable baseline. The scoring engine's *detection* is reliable; its *numeric scoring* is noisier than it should be at the category margins.

## Proposed approach

Four prompt-level (SKILL.md) changes, highest-leverage first. The goal is to pull *undue* variance (the 0.55-class swings) back toward the inherent ±0.3, **without** introducing false precision or distorting the finding set.

[...]

PR #51, opened 2026-05-30
test(baselines): re-freeze on Opus 4.8 as v0.7.7-claude48 (median-of-3) + widen 5 bands

virtualsteve-exa:
## Re-baseline the regression suite on Opus 4.8 (`v0.7.7-claude48`) + a more robust freeze process

Re-freezes the eleven-target regression baseline on the **unchanged v0.7.7 skill** under **Claude Opus 4.8**. `SKILL.md`, `schema.py`, `render.py`, `manifest_to_findings.py`, and the four knowledge bases are **byte-identical to v0.7.7** — the only variables are the reference model (4.7 → 4.8) and the freeze method.

> **Stacked on #49.** Base is `chore/sync-dev-with-main` so the diff here is just the two substantive commits; GitHub auto-retargets this to `dev` once #49 (the dev catch-up) merges. **Merge #49 first.**

### The more robust freeze process (median-of-3)
A single run is a noisy way to freeze a baseline — parts of the analysis are LLM judgement, so one run's weighted score carries variance even though the finding set is stable. Each of the eleven targets was characterized over **three independent runs** (33 runs total), holding skill/sources/remits constant:
- **Committed exemplar per target = its median run** (one real, unedited run — keeps the byte-render gate honest).
- **Bands set from the 3-run mean ± observed spread**, not any single run.
- New `tests/README.md` section **"Re-baselining (multi-run characterization)"** documents this as the standard procedure.

### Findings
- Opus 4.8 carries a **mild +0.22 mean upward lean** in weighted RAISE, **inverted-U by maturity** (largest on mid-maturity production targets; ~zero/negative at the extremes — `openhands` actually −0.25).
- **No Critical theme dropped on any of the 33 runs.**
- Five bands updated to fit the 4.8 means; the other six unchanged:

| Target | was | now | reason |
|---|---|---|---|
| aider | 1.1–1.8 | **1.8–2.4** | stable offset (σ 0.087) → move |
| langchain-sql | 0.7–1.4 | **0.9–1.6** | stable mild offset |
| sweep | 0.7–1.3 | **0.9–1.4** | marginal |
| openai-customer-service | 0.6–1.3 | **0.7–1.7** | noisy/centred (σ 0.28, highest) → widen, don't move |
| openhands | 1.8–2.5 | **1.7–2.5** | floor lowered |

### Also in this PR
- `docs/understanding-variability.md` — operator-facing page on run-to-run variability (why the LLM synthesis stage varies, the deterministic render stage doesn't, how to re-analyze with the LLM, when to run multiple times). Linked from the docs index + Interpreting Reports + Challenging Findings.
- `v0.7.7-sequential` retired (4.7 reference), kept for diff archaeology; README pointers + OWASP coverage report re-pointed.
- Per-category scoring variance tracked separately in RFE #48; the deferred `as_uri()` hardening is #50.

### Validation
- `python3 tests/render/test_render.py` → **308 passed, 0 failed** (66 new checks for the `v0.7.7-claude48` set: schema-valid, byte-identical re-render, verbatim remit quotes).


virtualsteve-exa:
Superseded by #52 — auto-closed when the base branch (`chore/sync-dev-with-main`) was deleted on the #49 merge. The same two commits, rebased clean onto `dev`, are in #52.

PR #53, opened 2026-05-30
chore(ci): enforce dev-not-behind-main + document the realign rule

virtualsteve-exa:
## Enforce "`dev` not behind `main`" + document the realign rule

**Answers "how did `dev` get behind `main`, and what stops it recurring?"**

### What happened
The post-release realign of `dev` to `main` (`git reset --hard origin/main`) has been the documented practice since 0.7.2 — but it's a **manual step, and it was skipped after 0.7.6 and 0.7.7**. Because releases are squash-merged onto `main` from a `release/x.y` branch (cut from `dev`), the release-finalization content (version bump, CHANGELOG, baseline freeze, spec edits) landed only on `main` and never returned to `dev`. Two missed realigns = `dev` two releases behind (`0.7.5`-content while `main` shipped `0.7.7`). Surfaced when trying to PR the `v0.7.7-claude48` rebaseline into `dev`; reconciled via the #49 back-merge.

It wasn't a missing rule — it was a missing **safety net**. This PR adds one.

### Changes
- **`.github/workflows/branch-drift.yml`** — scheduled (weekdays) + `workflow_dispatch` check that **fails when `dev`'s `plugin.json` release version is older than `main`'s**. Version-aware on purpose: a raw `git diff main dev` would false-positive constantly, because `dev` is *supposed* to be ahead of `main` between releases and squash-merging means the two branches never share commit SHAs. The only real failure mode is `dev` sitting on an older release — which is exactly what this checks. Logic verified locally (semver-aware via `sort -V`; current `dev`==`main`==`0.7.7` → passes; a simulated `0.7.7` vs `0.7.8` → fails).
- **`CONTRIBUTING.md` → "Keeping `dev` in sync with `main`"** — documents the divergence mechanism and three things: the post-release `reset --hard` realign (primary), the `chore/sync-dev-with-main` back-merge (recovery once `dev` has diverged, as in #49), and the optional "finalize releases through `dev`" structural fix that makes the realign a no-op.

### Note
The drift workflow is `schedule` + `workflow_dispatch` only, so it doesn't gate this PR; happy to `workflow_dispatch` it once merged to confirm a green run on the real branches.

### DIFF
--- a/.github/workflows/branch-drift.yml
+++ b/.github/workflows/branch-drift.yml
@@ -1,0 +1,44 @@
+# Copyright 2026 Exabeam, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Fails if `dev` is on an OLDER release than `main` — i.e. a release (or hotfix)
+# landed on `main` and was never back-merged into `dev`. See CONTRIBUTING.md,
+# "Keeping `dev` in sync with `main`".
+#
+# This deliberately checks the release VERSION, not a raw `git diff`: `dev` is
+# expected to be *ahead* of `main` between releases (unreleased feature work),
+# and squash-merging means the two branches never share commit SHAs. The only
+# real failure is `dev` sitting on an older version than `main`.
+name: Branch drift
+
+on:
+  schedule:
+    - cron: "17 14 * * 1-5" # weekday mid-day UTC
+  workflow_dispatch: {}
+
+permissions:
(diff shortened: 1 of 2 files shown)


virtualsteve-exa:
Good catch — applied in 943b601: `git fetch origin &&` now leads the realign so the hard reset can't target a stale tracking ref. Resolved.

PR #57, opened 2026-05-31
skill(pre-flight): author remits from docs as intent, not from code as description

virtualsteve-exa:
### The bug
The Pre-flight remit-authoring path instructed: *"Fill each template section with what the source says."* When pointed at an agent's code, that produces a remit that **mirrors the implementation** — and a clause that narrates observed behavior can never be found *in violation*, so a scan against it finds **zero gaps**. It defeats the policy-vs-implementation comparison Praxen exists to make. (`PRAXEN_SPEC.md` §237 already said the right thing — *"the remit is a policy document, not a system description… it declares intent"* — the skill's Pre-flight had drifted from the spec.)

### The fix
- **Step 2 — source.** Author from the agent's **documentation** (README, `SECURITY.md`, `AGENTS.md`, design docs, the operator's description), **not its code**. Read source only when docs are *spectacularly unavailable*; when docs merely go *silent* on a section, write an `[Inferred]` inference or ask the operator — never code-spelunk. Checking code against the remit is the *scan's* job.
- **Step 3 — framing.** Intent-first: the **violability test** (could a non-compliant build violate this clause? if not, it's a description → rewrite or drop), **don't pre-authorize observed risk**, **invert adversarial/attack-only docs** (CTF write-up / pentest report) into intended-secure policy, **avoid the mirror** (including the author's own "it's just a heuristic" self-assessment), and **don't carry `[Verified]` tags** into the remit.

### Evidence (A/B, validated across 4 targets)
Holding the authoring prompt constant and changing only the skill barely moved the output (intent-*framing* alone: 60 `[Verified]` leaks, 2 MUST). Switching to docs-first **sourcing** was the lever:

| | `[Verified]` leaks | MUST/MUST-NOT | code citations |
|---|--:|--:|--:|
| code-prompt remit | 74 | 1 | 33 |
| **docs-first remit** | **0** | **36** | **0** |

Regenerating finbot / aider / openhands docs-first reproduced it: 0 leaks, ~0 code cites, real prescriptive obligations — with `[Inferred]` volume honestly tracking each agent's doc quality (finbot 51 → openhands 8).

### Safety / scope
**Scan-orthogonal** — Pre-flight runs only at *authoring* time; this does not change scan behavior, the schema, or the committed baselines. No `PRAXEN_SPEC` change needed (the spec already states the principle). Deferred follow-ups (tracked, not in this PR): template-completeness verbosity, and a calibrated assert-vs-flag pass to reduce over-flagging without re-introducing invented intent.


virtualsteve-exa:
Added a second remit-generator fix on this branch (commit 9b7d1b1): **drop the `[Inferred]` hedging and stop posing implementation questions as remit gaps.** The path was tagging clauses `[Inferred]` and asking "is X on by default? / how is Y stored?" — but those are facts about the *code* the scan discovers, not remit-time decisions. Now: every policy-body clause is a stated obligation (no `[Inferred]`/`[Verified]` tags), implementation uncertainty is stated as intent (and the scan finds the divergence), and genuine un-derivable *operator* questions (scope, thresholds, allowlists) go in a labeled 'Open Questions for the operator' section at the end of the file, below the footer — outside the scanned policy. Validated docs-first on finbot/aider/openhands: `[Inferred]` in the policy body **51/28/8 → 0/0/0**, obligations held, operator questions externalized. Both fixes are scan-orthogonal (authoring-time only).

virtualsteve-exa:
Both inline comments were on the `[Inferred]` handling in Step 3, which later commits on this branch **removed entirely** — the path no longer tags clauses `[Inferred]` at all (the "one exception is [Inferred]" line and the inferred-clause bullet are both gone; the rule is now stated positively: every clause is a stated obligation, uncertainty goes to a separate 'Open Questions for the operator' section). So both are superseded — resolving. 'inferred' now lives only on the scan side.

PR #63, opened 2026-05-31
feat: add RAISE score distribution coverage report

virtualsteve-star:
## Summary

- Adds `tests/baselines/raise_coverage.py` — a new coverage script, structured identically to `owasp_coverage.py`, that renders a self-contained HTML report of RAISE scores across the full baseline set
- Adds `tests/baselines/CURRENT` — an explicit pointer to the canonical baseline dir (`v0.7.7-claude48`), read by both coverage scripts; eliminates mtime-based selection which is nondeterministic on CI
- Pre-generates `tests/baselines/raise-coverage-report.html` from `v0.7.7-claude48` (12 targets); the GitHub Pages link goes live once this reaches `main` via the normal dev→main promotion PR
- Removes the `release.yml` auto-commit-to-main steps (incompatible with branch protection); coverage reports are committed on `dev` as part of release prep and reach `main` through the normal promotion flow, with a comment in the workflow documenting the manual regen commands
- Updates default baseline detection in both scripts to use `CURRENT` first, falling back to lexicographic sort of `v*/` dirs
- Adds a live-report callout to `docs/RAISE.md` linking to the GitHub Pages URL, parallel to the existing callout in `docs/owasp.md`

## What the report shows

Three sections:

1. **Per-target RAISE score heatmap** — all baseline targets as rows, the six RAISE categories + weighted overall as colour-coded columns (red = Absent, teal = Exemplary), sorted by weighted overall descending
2. **Per-component score distributions** — for each of the six categories, how many baseline targets landed at each maturity level (0–5), with population average and standard deviation
3. **Population weighted-overall histogram** — the full population bucketed by maturity band

## Test plan

- [ ] Confirm `python3 tests/baselines/raise_coverage.py` runs cleanly against the current baseline set (stdlib-only, no pip install required)
- [ ] Confirm `python3 tests/baselines/owasp_coverage.py` still runs cleanly (default baseline detection change)
- [ ] Open `tests/baselines/raise-coverage-report.html` in a browser and verify heatmap, distribution charts, and histogram render correctly
- [ ] Verify `docs/RAISE.md` link points to the correct GitHub Pages URL after this reaches `main`


virtualsteve-exa:
## Review — held on three high-severity items

Solid script overall: stdlib-only, mirrors `owasp_coverage.py` cleanly, and both scripts ran green when I applied the patch. But there are three blockers, two confirmed by running the code and by querying this repo's actual branch protection.

### 🔴 Blocking

**1. The `release.yml` auto-push to `main` cannot succeed — and would break the branch invariant if it did.**
The new step does `git push origin HEAD:main` with a `[skip ci]` commit. Live protection on `main`:
```
required_reviews: true   (Require a PR before merging — blocks ALL direct pushes)
required_status_checks: strict, contexts = dco, test(3.9/3.12/3.13)
restrictions: false   enforce_admins: false
```
- "Require a pull request before merging" is **ON**, so a direct push from `github-actions[bot]` (the default `GITHUB_TOKEN`, not an admin) is rejected outright. The strict required status checks compound it, and `[skip ci]` guarantees those checks never run, so the commit could never satisfy them.
- This push happens **after** `gh release create` already ran, so the job fails *mid-release* → half-completed releases (tag + GitHub release exist, reports never land).
- Even if it succeeded, it commits to `main` but never touches `dev`, so `main` carries a commit `dev` lacks — directly violating the **"main is an ancestor of dev"** invariant that `branch-drift.yml` just hardened (commits ab81d83 / a18c9b1). The next scheduled branch-drift run would fail until someone manually FF's dev.

Pages serves from `main` at `/`, so the reports do need to reach main — but auto-committing from the release job is incompatible with this repo's protection + branching model. Consider a Pages-artifact deploy, or regenerating the reports as part of the normal dev→main promotion PR rather than a post-release push.

**2. The committed `raise-coverage-report.html` was generated from the WRONG baseline.**
The PR body says it's pre-generated "from the current `v0.7.7-claude48` baseline." It isn't:

| | committed RAISE report | committed OWASP report | canonical (claude48) |
|---|---|---|---|
| baseline dir | `v0.7.7-sequential` | `v0.7.7-claude48` ✓ | `v0.7.7-claude48` |
| targets | **11** | 12 | 12 |
| pop. avg | **1.19** | — | 1.55 |
| Hermes (12th target, added in #58) | **absent** | present | present |

So the RAISE Pages report would ship the wrong population numbers and silently **omit the 12th baseline target that #58 just added**, while the OWASP report beside it shows the correct set. The per-target `report ↗` links also point into `v0.7.7-sequential/...`.

**3. Root cause of #2: mtime-based `_default_baseline()` is nondeterministic and intent-ambiguous.**
```python
sorted(THIS_DIR.glob("v*/"), key=lambda p: p.stat().st_mtime, reverse=True)
```
Git doesn't store mtimes — a fresh `actions/checkout` writes all `v*/` dirs at ~checkout time in arbitrary order, so the release-time regeneration could build from *any* version (even `v0.4-parallel`). It also can't tell `v0.7.7-claude48` from `v0.7.7-sequential` by intent — whichever was touched last wins. Reproduced locally: `touch v0.7.7-sequential` and the auto-pick immediately switched to it — exactly how the wrong artifact in #2 got produced. **Recommend** an explicit canonical pointer (a `CANONICAL = "v0.7.7-claude48"` constant, or a small `current-baseline` marker file) rather than mtime; a plain version sort is still ambiguous across the `-claude48`/`-sequential` split.

### 🟡 Should fix
- **"Live immediately on merge" is inaccurate.** Pages serves from `main`; this PR's base is `dev`. The report only goes live once it reaches `main` (next promotion/release), not on merge to dev.

### Nits
- `git apply` flags 5 trailing-whitespace lines in the new files.
- Extra bare blank line added near `docs/RAISE.md:11` (harmless).

### What's good
- Script structure mirrors `owasp_coverage.py`; clean `None`-handling; RAISE weights sum to 1.0; JSON mapping (`raise_posture.weighted_overall`, `categories[].key/score`) matches the baseline schema; both scripts ran cleanly.
- The `owasp_coverage.py` `DEFAULT_OUT` change (`Path.cwd()` → `THIS_DIR`) is correct and necessary.

**Bottom line:** the generator is good; the **delivery mechanism (release.yml → main)** and the **mtime baseline selection** are the blockers, and the latter already corrupted the committed artifact. Regenerate against `v0.7.7-claude48`, replace mtime selection with an explicit canonical pointer, and rethink how reports reach `main` given branch protection.

virtualsteve-star:
Thanks for the thorough review — all three blockers, the should-fix, and the inline suggestions are addressed in commit 1a7a62e. Summary of what changed:

## Blockers

**1. release.yml auto-push to main**
Removed both new steps (`Regenerate baseline coverage reports` and `Commit and push updated coverage reports`) entirely. Replaced with a comment in the workflow pointing to the manual regeneration commands. Coverage reports now follow the same pattern as before this PR: committed on `dev` as part of release prep, reaching `main` through the normal promotion PR.

**2. Committed HTML built from wrong baseline**
The `raise-coverage-report.html` was built from `v0.7.7-sequential` (11 targets) rather than `v0.7.7-claude48` (12 targets) — Hermes was absent and population numbers were wrong. Regenerated from the correct baseline; the header now correctly shows `v0.7.7-claude48` and all 12 targets.

**3. mtime-based `_default_baseline()` is nondeterministic**
Replaced the mtime sort in both scripts with an explicit `CURRENT` pointer file at `tests/baselines/CURRENT` (contains `v0.7.7-claude48`). Both scripts now:
1. Read `CURRENT` and use the named dir if it exists
2. Fall back to a lexicographic sort of `v*/` dirs (version strings sort correctly without mtimes)

This is deterministic on fresh CI checkouts and makes the canonical baseline an explicit, reviewable file rather than an implicit runtime state. Updating the pointer when a new baseline is added is one line in the release prep PR.

## Should fix

**"Live immediately on merge" wording** — corrected in the fix commit message; the PR description has also been updated below.

## Nits

- Extra blank line after `---` in `docs/RAISE.md`: removed.
- Trailing whitespace: grep found none in the current working tree (the edits had already cleaned the lines flagged by `git apply`).

## Gemini inline suggestions

Both accepted:
- `glob("v*/")` → `[p for p in glob("v*") if p.is_dir()]` in both scripts (applied to all three call sites).
- `try/except (json.JSONDecodeError, OSError)` added around JSON loading in `raise_coverage.py`'s `gather()`, with a stderr warning and `continue` so one corrupt file doesn't abort the whole report.

virtualsteve-exa:
## Re-review of `1a7a62e` — all three blockers verified fixed ✅

Pulled the fixup commit and re-ran everything against the local baselines.

**1. release.yml auto-push to main** — ✅ Removed entirely. The post-release commit/push step is gone; reports now reach `main` via the normal dev→main promotion PR, with a comment documenting the manual regen commands. No more half-release failure mode, no invariant violation.

**2. Wrong committed baseline** — ✅ `raise-coverage-report.html` now renders from `v0.7.7-claude48`: **12 targets**, **pop. avg 1.55**, **Hermes present**, and the `report ↗` links point into `v0.7.7-claude48/...`. Matches the OWASP report's baseline.

**3. mtime selection** — ✅ Replaced with a `tests/baselines/CURRENT` pointer (`v0.7.7-claude48`), read first by both scripts with a name-sorted fallback. Verified robust: with `v0.7.7-sequential` `touch`'d to the newest mtime, both scripts still correctly pick `v0.7.7-claude48`. (Bonus: nice defensive try/except around `json.load` for malformed findings.)

Nits also clear — no trailing whitespace in the updated scripts.

One leftover from the original review, non-blocking: the PR summary's "GitHub Pages link is live immediately on merge" is still slightly off (Pages serves from `main`; this merges to `dev`, so it goes live on the next promotion). Worth a one-line tweak to the description, but doesn't block.

LGTM from my side — leaving the merge call to Steve.

virtualsteve-star:
PR description updated — replaced "live immediately on merge" with "goes live once this reaches `main` via the normal dev→main promotion PR", and rewrote the summary bullets to reflect the fixes (CURRENT file, removed release.yml steps). The test-plan wording is updated to match.

Issue #65, opened 2026-05-31
Tool feedback: framework scanning, manifest resilience, and discovery gaps (from uAgents scan)


virtualsteve-star:
## Context

This is practitioner feedback from running Praxen against `fetchai/uAgents` (see #64 for the baseline RFE). Two full scans were completed across this session and the prior one (`maeste/multi-agent-a2a`). The feedback below is grounded in what actually happened, not hypotheticals.

---

## 1. No recovery point before Step 9.9

**What happened:** Steps 1–8 involve 25–30 file reads on a large codebase (uAgents has 80+ Python files across four packages) with no artifact written to disk until the Step 9.9 manifest. The uAgents scan nearly exhausted the context window during evidence gathering, before any checkpoint existed.

**What would help:** A lightweight intermediate checkpoint after Step 4 — even a flat text file listing files read and first-pass signals (credential patterns found, binding addresses noted, dependency pins observed). It doesn't need to be machine-parseable. Its only job is to survive a compaction so the analysis pass doesn't re-read from scratch.

**Suggested location:** `./reports/<slug>-evidence-<TIMESTAMP>.txt`, written at the end of Step 4 before Step 4b.

---

## 2. Manifest format failure modes are silent until parse time

**What happened:** The manifest format has correct but subtle rules — depth-0 fields, depth-2 array items, depth-4 continuations, null vs. omit conventions, block prose vs. flat bullets. Getting one wrong produces a parser error, not a hint. Debugging format late in a long scan is expensive.

**What would help:** A `--validate-manifest` flag on `manifest_to_findings.py` that checks structure and reports all errors without emitting JSON. Running this after the skeleton write (before appending rules and findings) would catch format problems early.

**Secondary ask:** Document the most common errors with examples in the manifest template comment block, not just in the skill file. Practitioners who've read the skill once and are mid-scan shouldn't need to re-read 200 lines of documentation to debug indentation.

---

## 3. 25-word summary cap breaks down for compound or nuanced findings

**What happened:** The predictable seed fallback finding in uAgents (A2A adapters use `f"{name}_seed"` as a fallback — anyone who knows the adapter name can reconstruct the private signing key) is genuinely Critical but requires more than 25 words to state accurately without losing the "why Critical" context.

The cap is a good discipline for simple findings. It's wrong for compound ones.

[...]

PR #66, opened 2026-06-01
feat(skill): scan-quality fixes from uAgents feedback (#65 items 4, 6, 7)

virtualsteve-exa:
## Summary

Three grounded, low-risk improvements from the `fetchai/uAgents` and `maeste/multi-agent-a2a` scans — the "wheat" from the #65 backlog. All **guidance-only**: no schema, version, or `PRAXEN_SPEC` change (verified the spec doesn't mirror either edited section), and the render gate stays green (314 passed / 0 failed).

| #65 item | Change | Why |
|----------|--------|-----|
| **4** | Add an **Infrastructure / deployment (IaC)** row to the Step 4 discovery table | The top Critical in the uAgents scan — a default `seed:` shipped in a committed Helm `values.yaml` — was nearly missed because nothing directed the scan into the deployment tree. `*.yaml` was already globbed; the *directive to go look* in `deploy*/helm/charts/k8s/manifests/terraform/infra` wasn't. This is a guidance gap, not a glob gap. |
| **6** | Promote the **code-first remit warning** to a callout at the top of the Pre-flight section | Authoring a remit from the code the scan will judge collapses the comparison → zero findings, the most common way to produce a useless scan. The guidance existed but was buried in paragraph 2; now it's the first thing an author sees. |
| **7** | Add **property-vs-mechanism** guidance to the template's Action Boundaries callout | Rules should state the required *outcome* (`endpoints MUST enforce caller authentication`) not the *technique* (`MUST NOT bind to 0.0.0.0`). Mechanism rules false-positive on correct container deployments and miss the real gap. |

## Deliberately deferred (the chaff)
From #65, **not** in this PR: #1 (Step 4 evidence checkpoint) and #5 (`scan_type: framework`) are design-heavy — separate PRs; #2 (`--validate-manifest`) and #3 (word-cap / two-field finding) are author-convenience / cosmetic and don't change finding quality.

## Test plan
- [x] `python3 tests/render/test_render.py` → 314 passed, 0 failed
- [ ] Sanity-check the rendered SKILL Step 4 table and Pre-flight callout
- [ ] Confirm the template's Action Boundaries guidance reads cleanly

Closes #65 items 4, 6, 7 (remaining items tracked separately).


virtualsteve-exa:
Closing without merging — on reflection these three edits are guidance nudges on a prompt that already says most of this, and the marginal behavior change doesn't justify the added prose:

- **#6** duplicates the code-first warning already in the Pre-flight section's item 2.
- **#4** restates "look harder for secrets" — `*.yaml` is already globbed and the skill already warns against skipping benign-looking files. The real IaC concern is better served by a regression *test* (fail when a scan misses a planted deployment secret) than a table row.
- **#7** (property-vs-mechanism) was the only genuinely distinct idea, but not worth a PR on its own.

Net: prose you can't cheaply measure is mostly faith, and every paragraph dilutes the ones that matter. Backing out. Tracking the disposition on #65.

PR #71, opened 2026-06-01
GitHub Pages landing page + shared design system

virtualsteve-exa:
Replaces the default README-rendered GitHub Pages landing with a custom, gorgeous HTML landing page, and unifies the visual theme across all three Pages surfaces.

## What's in here

**1. Landing page (`index.html`)** — served from the repo root on `main` (no Jekyll front matter → copied verbatim, takes precedence over `README.md`; `docs/*.md` still render).
- Hero with Praxy mascot, value prop, primary CTAs
- "What it catches" detection-pattern checklist
- **Scoring** section: FinBot example report + live OWASP and RAISE summary cards
- Install terminal with copy-to-clipboard marketplace commands
- Sponsor band using the **official white Exabeam logo** (brand-guide compliant)
- New transparent Praxen logo in nav/footer; web-optimized image assets in `graphics/web/`
- Sticky-nav-aware anchor scrolling; footer "Apache 2.0 License" links the repo `LICENSE`

**2. Shared design system (`assets/praxen-theme.css`)** — single source of truth (tokens, fonts, base, buttons, report chrome). Edit it once, all three surfaces retheme.
- `index.html` **links** it
- `owasp_coverage.py` / `raise_coverage.py` **inline** it at render time, so each coverage report stays a self-contained single file

**3. Dark-themed coverage reports** — both OWASP and RAISE reports restyled to match the landing page (Inter + Space Grotesk, orange accents). RAISE maturity heatmap keeps its red→cyan scale. **No data/logic changes** to the generators — same targets, gather, and scores; regenerated against `CURRENT` (`v0.7.7-claude48`).

## Verification
- `python3 tests/render/test_render.py` → **314 passed, 0 failed**
- `bash build.sh` → clean (exit 0)
- Both coverage generators compile and regenerate cleanly; zero leftover light-theme color references
- HTML validated; all internal doc/report links resolve on Pages (200)

## Notes
- Goes live on Pages only after `dev` → `main` promotion.
- The `v0.7.9-pre` baseline work is intentionally **not** here — it's parked on `wip/baseline-v0.7.9-pre` for the next tool rev.


index.html
```diff
@@ -0,0 +1,556 @@
...
+    (function () {
+      var btn = document.getElementById('copyBtn');
+      var ok = document.getElementById('copyOk');
+      var cmds = [
+        '/plugin marketplace add open-agent-ai-security/praxen',
+        '/plugin install praxen@open-agent-ai-security'
+      ].join('\n');
+      if (!btn) return;
+      btn.addEventListener('click', function () {
+        navigator.clipboard.writeText(cmds).then(function () {
```
virtualsteve-star: **Missing `.catch()` on clipboard Promise**

`navigator.clipboard.writeText()` returns a Promise but there is no rejection handler. When the Clipboard API is unavailable — non-HTTPS origin (`file://` or `http://`), Firefox private-browsing, or permission denied — the Promise rejects silently. The `ok` confirmation span never shows, and the browser logs an unhandled rejection in the console. The user assumes the copy worked and pastes nothing.

Suggested fix:
```js
navigator.clipboard.writeText(cmds).then(function () {
  ok.classList.add('show');
  setTimeout(function () { ok.classList.remove('show'); }, 1800);
}).catch(function () {
  ok.textContent = 'Copy failed — paste manually';
  ok.classList.add('show');
  setTimeout(function () {
    ok.classList.remove('show');
    ok.textContent = 'Copied ✓';
  }, 2500);
});
```

assets/praxen-theme.css
```diff
@@ -0,0 +1,134 @@
...
+
+  Structure:
+    1. Fonts (@import — must stay first so inlined <style> blocks remain valid)
+    2. :root design tokens
+    3. Global base + primitives (used everywhere)
+    4. .report-page — chrome shared by the coverage reports only (scoped so it
+       never leaks onto the landing page, which does not carry that body class)
+*/
+
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=Space+Grotesk:wght@500;600;700&display=swap');
```
virtualsteve-star: **Google Fonts loaded twice on the landing page**

`index.html` also has an explicit `<link href="https://fonts.googleapis.com/..."\>` tag (line 31) loading the identical URL. The `<link>` tag fires first (render-blocking); this `@import` fires a second request for the same resource when the browser parses the linked CSS file.

Modern browsers deduplicate warm-cache hits, but cold loads (first visit, CI preview) issue two round-trips. More importantly, this splits ownership: a future font weight/family change must be applied in both places, and they will eventually diverge.

The fix is to remove one of the two. Since the coverage reports rely on the `@import` for the font (they inline the CSS and have no separate `<link>` tag), keep the `@import` here and remove the redundant `<link>` from `index.html`.

tests/baselines/owasp_coverage.py
```diff
@@ -49,6 +54,20 @@ def _default_baseline() -> Path:
 DEFAULT_BASELINE = _default_baseline()
 DEFAULT_OUT = THIS_DIR / "owasp-coverage-report.html"
 
+
+def load_theme_css() -> str:
```
virtualsteve-star: **`load_theme_css()` is copy-pasted identically into both coverage scripts**

The same function (path computation, read, OSError handler) also appears at line 61 of `raise_coverage.py`. The only difference is the script name in the warning string.

Since there is no existing shared utility module for these scripts, a minimal fix is a single `tests/baselines/theme_utils.py` (or similar) that both import. Any future change to the path logic or error message then only needs one edit.

Not a blocker for this PR — the duplication is stable and the two copies are currently identical — but worth a follow-up before they drift.

tests/baselines/raise_coverage.py
```diff
@@ -52,6 +57,20 @@ def _default_baseline() -> Path:
 DEFAULT_BASELINE = _default_baseline()
 DEFAULT_OUT = THIS_DIR / "raise-coverage-report.html"
 
+
+def load_theme_css() -> str:
```
virtualsteve-star: **Duplicate of `load_theme_css()` in `owasp_coverage.py`**

See comment on `owasp_coverage.py` line 58. Both copies are currently identical; a shared helper would remove the sync burden.

virtualsteve-star (commented):
Multi-angle code review (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude). Five findings; no blockers, all fixable. Details inline.

virtualsteve-star:
**Committed generated HTML artifacts go stale silently** (`owasp-coverage-report.html`, `raise-coverage-report.html`)

The two pre-generated HTML files are committed alongside the generators that produce them. A follow-on PR that renames a CSS token in `assets/praxen-theme.css` or adjusts category logic in the generators — but skips the regeneration step — will leave the committed files diverged from what the generators now produce. Anyone linking to or viewing them in-repo sees outdated output with no indication it is stale.

This is an architectural note rather than a bug in this PR (which correctly regenerates both files). Options to consider:
1. **CI guard**: add a step that re-runs both generators and fails if the output differs from the committed file (via `git diff --exit-code`).
2. **Do not commit artifacts**: serve the reports as CI artefacts or generate them at Pages deploy time, removing the need to keep the committed copies in sync.

Either way, the current approach works fine as long as the regeneration discipline holds — but a CI check would make that discipline automatic.

index.html
```diff
@@ -0,0 +1,556 @@
...
+          ok.classList.add('show');
+          setTimeout(function () { ok.classList.remove('show'); }, 1800);
+        });
+      });
```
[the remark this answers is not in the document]
virtualsteve-exa: Fixed in 62898f5 — the handler now guards `navigator.clipboard`, adds a `.catch()`, and falls back to a hidden `<textarea>` + `document.execCommand('copy')` for non-secure contexts, with a "Copy failed — paste manually" message if even that fails.

index.html
virtualsteve-exa: Fixed in 62898f5 — added `.catch()`. On rejection (or no Clipboard API) it falls back to `execCommand('copy')`; if that also fails the confirmation span shows "Copy failed — paste manually" for 2.5s, so a failed copy is never silent.

assets/praxen-theme.css
virtualsteve-exa: Fixed in 62898f5 — removed the redundant `<link rel="stylesheet">` from index.html. The shared theme's `@import` is now the single owner of the font load (the coverage reports need it since they inline the CSS). Kept the two `preconnect` hints, which still warm the `@import` fetch.

tests/baselines/owasp_coverage.py
virtualsteve-exa: Done in 62898f5 — folded the follow-up in: extracted `load_theme_css()` (and the path computation) into `tests/baselines/theme_utils.py`, now imported by both generators. (Caught and fixed a path bug while doing it — `parents[2]` from the file, not `parents[1]`.)

tests/baselines/raise_coverage.py
virtualsteve-exa: Resolved with the shared `tests/baselines/theme_utils.py` — see the reply on owasp_coverage.py.

index.html
```diff
@@ -0,0 +1,584 @@
...
+        if (navigator.clipboard && navigator.clipboard.writeText) {
+          navigator.clipboard.writeText(cmds)
+            .then(function () { flash(); })
+            .catch(fallbackCopy);
+        } else {
+          fallbackCopy();
+        }
+      });
+    })();
+  </script>
```
virtualsteve-star, line 582: **Clipboard fix: LGTM**

The `flash()` helper avoids repetition cleanly. The three-tier path (async Clipboard API → `execCommand` fallback → visible failure message) covers the realistic failure modes. A few observations for the record:

- `ta.style.opacity = '0'` is fine for desktop; on iOS Safari `execCommand` occasionally requires a visible, on-screen element. For this use case (a GitHub Pages docs page viewed on a modern browser) it's acceptable — just noting it if anyone ever files a Safari-mobile bug.
- `flash()` without arguments does not reset `ok.textContent` before showing it, so if a previous failure message is still displayed when the success path fires (unlikely — requires a second click mid-timeout), the user would briefly see the old failure text. Extremely edge-case; not worth changing.

index.html
```diff
@@ -0,0 +1,584 @@
...
+  <meta property="og:title" content="Praxen — Make sure your agent does its job, and only its job.">
+  <meta property="og:description" content="Open-source Agent Behavior Verification. Compare declared intent to real-world evidence and catch behavior drift before it becomes a risk.">
+  <meta property="og:url" content="https://open-agent-ai-security.github.io/praxen/">
+  <meta property="og:image" content="https://open-agent-ai-security.github.io/praxen/graphics/praxen-social.png">
+  <meta name="twitter:card" content="summary_large_image">
+  <meta name="twitter:title" content="Praxen — AI Agent Behavior Verifier">
+  <meta name="twitter:description" content="Make sure your agent does its job, and only its job.">
+  <meta name="twitter:image" content="https://open-agent-ai-security.github.io/praxen/graphics/praxen-social.png">
+
+  <!-- Preconnect warms the @import font fetch inside the shared theme below. -->
```
virtualsteve-star, line 29: **Font deduplication fix: LGTM**

Redundant `<link>` removed, preconnect hints kept with an explanatory comment. The `@import` inside the shared theme is now the single owner of the font load, which is the right call — the coverage reports inline the CSS and have no separate `<link>` to fall back on.

tests/baselines/theme_utils.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2026 Exabeam, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""Shared helpers for the Praxen baseline coverage-report generators
+(owasp_coverage.py and raise_coverage.py)."""
+import sys
+from pathlib import Path
+
+# tests/baselines/theme_utils.py -> repo root -> assets/praxen-theme.css
+THEME_CSS = Path(__file__).resolve().parents[2] / "assets" / "praxen-theme.css"
```
virtualsteve-star, line 9: **Path arithmetic verified: `parents[2]` is correct**

`Path(__file__).resolve()` on `tests/baselines/theme_utils.py` gives the file path, so:
- `parents[0]` = `tests/baselines`
- `parents[1]` = `tests`
- `parents[2]` = repo root ✓

The original scripts used `Path(__file__).resolve().parent` first (yielding the directory `tests/baselines/`), then `.parents[1]` from that — same destination. The refactor would have been wrong at `parents[1]` from the file; the dev caught and corrected that during the extraction. Good catch.

virtualsteve-star (commented):
Follow-up review of commit 62898f5. All four inline findings addressed — details on each thread. One minor nit remains (see inline on `praxen-theme.css`). The refactor also caught a latent path bug, which is a nice bonus.

assets/praxen-theme.css
virtualsteve-star: **Nit (unaddressed from first review): `@import` follows a 21-line block comment**

The CSS spec permits block comments before `@import`, so this is correct in all conforming browsers. It only bites non-spec CSS processors (headless HTML-to-PDF renderers, some email clients, strict minifiers) that treat any pre-`@import` content as a rule and silently drop the import — fonts fall back to `system-ui` with no error.

Completely fine to leave as-is for a GitHub Pages deployment. If coverage reports are ever piped through a preprocessor, consider moving the comment block after the `@import`.

virtualsteve-star:
**Follow-up review of `62898f5` — all four inline findings addressed**

**Clipboard (`index.html`)** — clean three-tier implementation: async Clipboard API → `execCommand` textarea fallback → visible "Copy failed — paste manually" message. The `flash()` helper avoids repetition. Two micro-nits noted inline (iOS Safari edge case, mid-timeout double-click) but neither warrants a change.

**Font double-load (`index.html`)** — redundant `<link rel="stylesheet">` removed, `preconnect` hints kept with an explanatory comment. The `@import` inside the shared theme is now the single owner of the font load.

**`load_theme_css()` duplication (`owasp_coverage.py`, `raise_coverage.py`)** — cleanly extracted to `tests/baselines/theme_utils.py`, imported by both generators. Verified: `from theme_utils import` works correctly when the scripts are run as scripts (Python adds the script directory to `sys.path`).

**Path arithmetic (`theme_utils.py:9`)** — `Path(__file__).resolve().parents[2]` is correct for `tests/baselines/theme_utils.py` → repo root. The dev caught a real latent bug during the refactor: using `parents[1]` from the file path would have resolved to `tests/` instead of the repo root. Good catch.

**One nit still open:** the `@import`-after-comment block in `assets/praxen-theme.css:23` (noted inline). Harmless for a browser/Pages deployment; only bites non-spec CSS processors. Fine to leave as-is.

assets/praxen-theme.css
virtualsteve-exa: Done in c3f8425 — moved the `@import` to line 1, ahead of the header comment, so it's the very first token both in the file and when inlined into the reports' `<style>`. Took the fix precisely because the inlined coverage reports are the artifacts most likely to be piped through a PDF/minifier. Regenerated both reports.

PR #74, opened 2026-06-01
Styled GitHub Pages docs site (guide/) generated from docs/*.md

virtualsteve-exa:
Turns the user-facing docs into an on-brand HTML site instead of the default Jekyll markdown rendering, while keeping the raw markdown as the dev-facing / distributable source.

## Approach
`docs_build.py` (build-only) converts each `docs/*.md` → a self-contained styled page under `guide/`, **inlining `assets/praxen-theme.css`** — the same pattern the coverage-report generators use. Edit docs in markdown; regenerate at release.

Each page has:
- **Left-nav** listing every doc; the current one expands into its H2 section TOC
- Styled **GFM tables, fenced code, blockquotes, lists** on the dark theme
- **Rewritten cross-links**: sibling `foo.md` → `foo.html`; `../repo/file` → GitHub blob view; external/anchor links untouched
- An "Edit this page on GitHub" link back to the source `.md`

## Wiring
- New **`.docs-page`** block in `praxen-theme.css` (layout, nav, prose typography) — one theme still drives landing + reports + docs
- Landing page doc links repointed `docs/*.html` → `guide/*.html`
- **`requirements-dev.txt`** pins the build-only `markdown` dep. The shipped skill stays stdlib-only; `build.sh`'s allowlist already ships `docs/*.md` and excludes `guide/` / `docs_build.py` / `assets/` (verified against the built zip)
- **CONTRIBUTING**: regenerate `guide/` when `docs/*.md` change

## Verification
- `python3 tests/render/test_render.py` → **314 passed, 0 failed**
- `bash build.sh` → distro contains `docs/*.md`, excludes all Pages-only artifacts
- Rendered spot-checks (usage = tables, quickstart = code) look on-brand

## Notes
- The raw `docs/*.md` are untouched (GitHub repo view + distro).
- A CI "docs are stale vs. their .md" check is a reasonable follow-up, not in this PR.


docs_build.py
```diff
@@ -0,0 +1,167 @@
...
+
+def onpage_toc(toc_tokens) -> str:
+    """Build a flat 'sections' list from the page's H2 headings."""
+    items = [t for t in toc_tokens if t.get("level") == 2]
+    if not items:
+        return ""
+    lis = "".join(
+        f'<li><a href="#{t["id"]}">{html.escape(t["name"])}</a></li>' for t in items
+    )
+    return f'<ul class="docs-subnav">{lis}</ul>'
```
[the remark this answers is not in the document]
virtualsteve-exa: Good catch — confirmed the section subnav was missing from every page (`toc_tokens` is hierarchical: top level was just the H1, with H2s under its `children`). Fixed in 6f3a43d by recursively collecting all level-2 tokens; regenerated guide/ and the on-page section nav now renders under the active doc.

docs_build.py
```diff
@@ -0,0 +1,167 @@
...
+      - ../repo/file: GitHub blob view (these are dev artifacts, not on the site)
+      - sibling docs foo.md[#frag]: foo.html[#frag]
+    """
+    def repl(m):
+        href = m.group(1)
+        if href.startswith(("http://", "https://", "mailto:", "#")):
+            return m.group(0)
+        if href.startswith("../"):
+            return f'href="{REPO}/blob/main/{href[3:]}"'
+        return 'href="' + re.sub(r"\.md(#|$)", r".html\1", href) + '"'
```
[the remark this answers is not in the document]
virtualsteve-exa: Fixed in 6f3a43d — `rewrite_links` now treats a link as external when `urlparse(href).scheme` is set (or it's protocol-relative `//`), instead of the hardcoded prefix list, so ftp/file/etc. links are left untouched.

docs_build.py
```diff
@@ -0,0 +1,167 @@
...
+
+    for src, label in PAGES:
+        text = (DOCS_DIR / src).read_text(encoding="utf-8")
+        text = LEADING_COMMENT.sub("", text, count=1)  # drop the license header comment
+        md.reset()
+        body = md.convert(text)
+        body = rewrite_links(body)
+
+        m = re.search(r"<h1[^>]*>(.*?)</h1>", body, re.DOTALL)
+        title = re.sub(r"<[^>]+>", "", m.group(1)).strip() if m else label
```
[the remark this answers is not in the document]
virtualsteve-exa: Fixed in 6f3a43d — the extracted H1 is `html.unescape`d before `page_html` re-escapes it, so titles with entities no longer double-escape in the <title>.

requirements-dev.txt
```diff
@@ -0,0 +1,8 @@
+# Build-only dependencies — NOT used by the shipped Praxen skill (which is
+# stdlib-only and requires no pip install). These are needed only to regenerate
+# the GitHub Pages site locally / at release time.
+#
+#   pip install -r requirements-dev.txt
+#
+# Used by docs_build.py to convert docs/*.md into the styled guide/ HTML.
+markdown>=3.5
```
virtualsteve-star: **Unbounded `markdown>=3.5` can cause the `build.sh` byte-diff freshness check to spuriously fail a release**

The `build.sh` freshness backstop works by regenerating `guide/` and running `git status --porcelain -- guide/` — a byte-level comparison against what was committed. Python-Markdown has version-to-version output differences (whitespace around block elements, HTML attribute quoting, entity choices) even across minor releases.

Concrete risk: a developer commits `guide/*.html` generated with `markdown==3.10.2` (currently installed). At release time CI does `pip install -r requirements-dev.txt`, gets `markdown==3.11.x` (or whatever is latest), regenerates with slightly different HTML, and `git status` is non-empty → `build.sh` exits 1. The release is blocked on a no-op recommit.

Fix: pin to a specific minor or narrow range, e.g.:
```
markdown>=3.10,<4
```
or lock to the exact version used for the committed artifacts.

docs_build.py
```diff
@@ -0,0 +1,183 @@
...
+from pathlib import Path
+from urllib.parse import urlparse
+
+try:
+    import markdown
+except ImportError:
+    sys.exit("docs_build.py: Python-Markdown not installed. "
+             "Run: pip install -r requirements-dev.txt")
+
+ROOT = Path(__file__).resolve().parent
```
virtualsteve-star, line 34: **`rewrite_links()` rewrites `href` attributes inside raw-HTML passthrough blocks, not just Markdown-generated links**

Python-Markdown passes raw block-level HTML through **unmodified** — no HTML escaping. Confirmed by test:

```python
md.convert('<div><a href="foo.md">link</a></div>')
# → '<div><a href="foo.md">link</a></div>'
```

So if any doc page contains raw HTML (outside a fenced code block) with an `.md` href — for example, to display a UI example or a note in a custom `<div>` — `rewrite_links()` will rewrite it. For actual navigation links in raw HTML this is correct behavior. The latent risk is a doc author writing a raw-HTML example showing an `.md` link that gets silently rewritten in the rendered output.

Fenced code blocks are safe: Python-Markdown HTML-escapes `"` to `&quot;` inside them, so the regex never matches. But raw HTML passthrough is not.

No current `docs/*.md` file triggers this — all docs are pure Markdown. Worth documenting in the script (a comment on the `rewrite_links` call or the function itself noting that raw HTML blocks are in scope) so future authors know to use fenced code for any href examples.

docs_build.py
```diff
@@ -0,0 +1,183 @@
...
+import sys
+from pathlib import Path
+from urllib.parse import urlparse
+
+try:
+    import markdown
+except ImportError:
+    sys.exit("docs_build.py: Python-Markdown not installed. "
+             "Run: pip install -r requirements-dev.txt")
+
```
virtualsteve-star, line 33: **`rewrite_links()` silently skips single-quoted `href` attributes in raw HTML**

The regex `r'href="([^"]+)"'` matches only double-quoted attributes. Valid HTML allows single quotes: `href='other.md'`. A raw-HTML block in a `.md` doc using single-quoted hrefs would not be rewritten to `.html`, leaving a broken link in the generated site.

Python-Markdown itself always emits double-quoted attributes, so Markdown-native links are safe. Only raw HTML authored with single quotes is affected. Currently latent (no such links exist). If coverage is wanted:

```python
return re.sub(r"href=(['\"])([^\1]+?)\1", repl_both_quotes, body)
```

docs_build.py
```diff
@@ -0,0 +1,183 @@
...
+ROOT = Path(__file__).resolve().parent
+DOCS_DIR = ROOT / "docs"
+OUT_DIR = ROOT / "guide"
+THEME_CSS = ROOT / "assets" / "praxen-theme.css"
```
virtualsteve-star, line 37: **`href[3:]` only strips one `../` — a `../../` link produces a broken GitHub URL**

```python
if href.startswith("../"):
    return f'href="{REPO}/blob/main/{href[3:]}"'
```

For `href="../../SECURITY.md"`, `href[3:]` gives `./SECURITY.md`, and the resulting URL is `.../blob/main/./SECURITY.md`. GitHub may resolve this but it is fragile.

In practice, docs in `docs/` are one level below the repo root, so `../` (one level) covers all legitimate cross-repo links. No current `docs/*.md` uses `../../`. Worth adding a comment or assertion to document the assumption:
```python
# docs/ is one level below root; ../path goes to the repo root.
# Deeper relative paths (../../) are not expected and would produce malformed URLs.
```

tests/baselines/owasp_coverage.py
```diff
@@ -54,6 +54,8 @@ def _default_baseline() -> Path:
 
 DEFAULT_BASELINE = _default_baseline()
 DEFAULT_OUT = THIS_DIR / "owasp-coverage-report.html"
+# Live styled docs (the guide/ site), linked from the report for further reading.
+DOCS_BASE = "https://open-agent-ai-security.github.io/praxen/guide"
```
virtualsteve-star: **`DOCS_BASE` is copy-pasted identically into both coverage generators; `theme_utils.py` is already imported by both**

```python
# owasp_coverage.py line 58
DOCS_BASE = "https://open-agent-ai-security.github.io/praxen/guide"

# raise_coverage.py line 61 — identical
DOCS_BASE = "https://open-agent-ai-security.github.io/praxen/guide"
```

The `theme_utils` module already exists as the shared helper for these two scripts (added in the previous PR). This constant belongs there. Same pattern as the `load_theme_css()` consolidation from last time: one copy gets updated when the URL changes, the other doesn't.

virtualsteve-star (commented):
Multi-angle review of the docs-build pipeline. Six findings; two actionable now, four latent. Details inline.

virtualsteve-star:
**PR #74 code review — docs-build pipeline**

Six findings; two worth fixing before merge, four latent notes.

**1. `requirements-dev.txt:8` — unbounded `markdown>=3.5` can spuriously fail the release** *(fix recommended)*
The `build.sh` freshness backstop does a byte-level `git status` diff between the committed `guide/*.html` and what `docs_build.py` regenerates in CI. Python-Markdown has minor version-to-version output differences (whitespace, entity encoding). If CI installs a newer minor version than what the developer used to commit the guide files, the diff is non-empty and the release exits 1 — blocked on a no-op recommit. Fix: pin to `markdown>=3.10,<4` (or to the exact version used to generate the committed artifacts).

**2. `docs_build.py:58` — `DOCS_BASE` copy-pasted into both coverage generators** *(fix recommended)*
Same constant in `owasp_coverage.py:58` and `raise_coverage.py:61`; `theme_utils.py` already exists and is imported by both — same consolidation as the `load_theme_css()` fix from PR #71.

**3. `docs_build.py:34` — `rewrite_links()` rewrites `href` in raw-HTML passthrough, not just Markdown links** *(latent, document it)*
Python-Markdown passes raw HTML blocks through unescaped. The `href="..."` regex matches those too. For navigation links in raw HTML, this is correct. For a doc page that uses a raw HTML block (not a fenced code block) to *display* an `.md` link as an example, the displayed href gets silently rewritten. Currently no `docs/*.md` triggers this — all docs are pure Markdown. Worth a comment on the `rewrite_links` call noting that raw HTML blocks are in scope.

**4. `docs_build.py:33` — single-quoted `href` attributes in raw HTML are not rewritten** *(latent)*
The regex only matches double-quoted attributes. `href='foo.md'` (single-quoted, valid HTML) would be left as `foo.md` in the rendered page. Python-Markdown always double-quotes its own output, so pure-Markdown links are safe.

**5. `docs_build.py:37` — `href[3:]` only strips one `../`; `../../` would produce a broken GitHub URL** *(latent, document the assumption)*
No current `docs/*.md` uses `../../`. Worth a comment asserting docs/ is one level below root.

**6. `docs_build.py` + `build.sh` — new `docs/*.md` without a `PAGES` entry is silently skipped, freshness check stays green** *(process note)*
The `extra` warning prints to stderr and exits 0, so CI never catches a new doc that was added to `docs/` but forgotten from `PAGES`. Acceptable behavior (documented), but worth noting: the freshness backstop only guards against *stale* pages, not *missing* ones.

requirements-dev.txt
virtualsteve-exa: Fixed in c9f7a62 — pinned `markdown==3.9` exactly (the version the committed guide/ was generated with), with a comment explaining the byte-diff backstop is the reason. Confirmed regenerating with 3.9 yields zero diff.

tests/baselines/owasp_coverage.py
virtualsteve-exa: Fixed in c9f7a62 — `DOCS_BASE` now lives in `theme_utils.py` and is imported by both generators, same as the `load_theme_css()` consolidation.

docs_build.py
virtualsteve-exa: Fixed in c9f7a62 — `rewrite_links` regex is now `href=(["'])(.*?)\1`, so single- and double-quoted hrefs are both handled. guide/ output is byte-identical (all current docs are double-quoted Markdown-native).

docs_build.py
virtualsteve-exa: Fixed in c9f7a62 — now strips all leading `../` runs (`re.sub(r'^(?:\.\./)+', '', href)`), so a stray `../../x` yields a repo-root path instead of `./x`. Kept your note about docs/ being one level below root as a comment.

docs_build.py
virtualsteve-exa: Documented in c9f7a62 — the `rewrite_links` docstring now notes it operates on raw-HTML passthrough blocks (correct for real nav links), and that an author showing a literal `.md` href as an example should use a fenced code block (where Python-Markdown escapes the quotes, so the regex never matches). No current doc triggers it.

requirements-dev.txt
```diff
@@ -0,0 +1,14 @@
...
+#   pip install -r requirements-dev.txt
+#
+# Used by docs_build.py to convert docs/*.md into the styled guide/ HTML.
+#
+# Pinned to an EXACT version on purpose: build.sh's freshness backstop compares
+# the regenerated guide/ byte-for-byte against the committed output, and
+# Python-Markdown's HTML can shift between releases (whitespace, attribute
+# quoting, entity choices). An exact pin keeps the committed artifacts and the
+# release-time regeneration in lockstep. Bump this and regenerate guide/ together.
+markdown==3.9
```
virtualsteve-star, line 14: **Pin version mismatch: `markdown==3.9` but the committed `guide/` was generated with `3.10.2`**

This commit pins to `markdown==3.9` but does **not** regenerate `guide/*.html`. Those files were committed in an earlier commit using the then-current `markdown==3.10.2` (confirmed: that is the version installed where they were generated, and `3.10.2` is the latest on PyPI).

At release time CI will:
1. `pip install markdown==3.9`
2. `python3 docs_build.py` — regenerates with 3.9
3. `git status --porcelain -- guide/` — if 3.9 and 3.10.2 produce any HTML difference → non-empty → `build.sh` exits 1 → release blocked

The fix to the unbounded-version finding is correct in principle, but the pin must match the version actually used to produce the committed artifacts. Either:
- Change the pin to `markdown==3.10.2` (the version that generated the files), **or**
- Keep `markdown==3.9`, install it locally, run `python3 docs_build.py`, and include the regenerated `guide/*.html` in this commit

docs_build.py
```diff
@@ -0,0 +1,193 @@
...
+    ("RAISE.md",                  "The RAISE Framework"),
+]
+
+LEADING_COMMENT = re.compile(r"^\s*<!--.*?-->\s*", re.DOTALL)
+
+
+def rewrite_links(body: str) -> str:
+    """Rewrite href targets in the generated HTML for the site:
+      - any URL with a scheme (http, https, mailto, ftp, …), a protocol-relative
+        //host, or a pure #anchor: left untouched
```
virtualsteve-star, line 60: **`rewrite_links` fix: LGTM**

All three issues addressed cleanly:
- Single-quoted hrefs: new regex `href=(['"])(.*?)\1` handles both quote styles ✓
- `../../` stripping: `re.sub(r"^(?:\.\./)+", "", href)` strips all leading `../` runs ✓
- Raw HTML passthrough: documented in the docstring ✓

One minor note on the new regex: `(.*?)` is lazy without `re.DOTALL`, so a multi-line `href` attribute value (extremely unusual in practice) would not match. Not a concern for real HTML.

tests/baselines/theme_utils.py
```diff
@@ -8,6 +8,10 @@
 # tests/baselines/theme_utils.py -> repo root -> assets/praxen-theme.css
 THEME_CSS = Path(__file__).resolve().parents[2] / "assets" / "praxen-theme.css"
 
+# Base URL of the live styled docs (the guide/ site); reports link here for
```
virtualsteve-star, line 11: **`DOCS_BASE` consolidation: LGTM**

Constant moved to `theme_utils.py` and both generators now import it from there. Same pattern as `load_theme_css()` from PR #71. ✓

virtualsteve-star (commented):
Follow-up review of c9f7a62. Three of the five inline findings are fixed cleanly; one introduces a new but related bug; one nit remains.

virtualsteve-star:
**Follow-up review of `c9f7a62` — one new bug, two fixes clean**

**🔴 New bug: `markdown==3.9` pin does not match the version used to generate the committed `guide/` files**

This commit pins `requirements-dev.txt` to `markdown==3.9` but does *not* regenerate `guide/*.html`. Those files were generated with `markdown==3.10.2` (the latest on PyPI, confirmed locally). At release time CI installs `3.9`, regenerates `guide/`, gets different HTML, and `git status --porcelain -- guide/` is non-empty → `build.sh` exits 1.

Fix: either change the pin to `markdown==3.10.2`, or keep `==3.9`, regenerate with it, and commit the updated `guide/` in the same changeset.

---

**✅ `rewrite_links()` — all three issues fixed**
Single-quoted hrefs handled, `../../` stripping corrected, raw-HTML passthrough documented. The new regex and rewrite logic are correct.

**✅ `DOCS_BASE` — consolidated into `theme_utils.py`**
Both generators import from the shared module. Clean fix.

requirements-dev.txt
virtualsteve-exa: Dug into this — the committed guide/ was generated with **markdown 3.9, not 3.10.2**, so the pin already matches. Evidence from this dev machine:

- It's **Python 3.9.6** (the project's declared floor). `markdown==3.10.2` is **not installable here** — `pip install markdown==3.10.2` → `ERROR: No matching distribution found` (3.10.2 requires Python ≥3.10). The newest installable is 3.9.
- Regenerating with the pinned 3.9 yields **zero diff** against the committed `guide/*.html` — i.e. the committed artifacts are 3.9 output.

So release CI (`pip install markdown==3.9` → `docs_build.py` → `git status --porcelain guide/`) reproduces the committed files byte-for-byte and the backstop passes.

There's also a correctness reason to stay on 3.9 beyond just matching: the project supports Python **3.9–3.13**, and the docs build can be run on the 3.9 floor (this machine). markdown 3.10.2 can't run on 3.9, so pinning it would break `docs_build.py` / `build.sh` for anyone on the floor Python. 3.9 is the newest release that works across the whole matrix.

If we do want to track the latest line, the bump has to happen on a Python ≥3.10 machine that regenerates `guide/` in the same commit (which the requirements-dev.txt note already calls out) — but it's optional, not a release blocker. Happy to do that as a follow-up if you'd prefer to move to 3.10.x.

PR #76, opened 2026-06-02
test+docs: gate manifest converter test in CI; tighten determinism & read-only claims

virtualsteve-exa:
## What

`tests/render/test_manifest_to_findings.py` exits 1 on `main` today, and CI never noticed because it only runs `test_render.py`. Reported by @emmanuelgjr — thank you.

### Root cause (verified)
The converter stamps `praxen_version` from `.claude-plugin/plugin.json` (`_load_praxen_version()`), now **0.7.8**. The committed golden `tests/fixtures/helperbot.from_manifest.json` was captured at **0.7.5** and never regenerated. The byte-comparison therefore fails on a single same-length field:

```
< "praxen_version": "0.7.5"   (golden)
> "praxen_version": "0.7.8"   (live converter output)
```

(Equal byte length is why the failure read `actual_len=33258 golden_len=33258` — `0.7.5`/`0.7.8` are both 5 bytes. The version line is the **only** delta; everything else is byte-identical.)

## Fix

1. **Gate it in CI.** Add `test_manifest_to_findings.py` to both `ci.yml` and `release.yml`. This was the real bug — a failing test across the manifest→findings half of the deterministic pipeline was hidden behind a green badge.
2. **Decouple the golden from the release version.** `praxen_version` is *injected from `plugin.json`*, not derived from the manifest, so it shouldn't couple the fixture to every version bump (which is how it lapsed silently 0.7.5 → 0.7.8). The round-trip now masks the version field and asserts the stamping behavior **directly** (`converter stamps current plugin.json version`). So the golden checks everything the manifest actually drives, and never goes stale on a bump again.
3. **Regenerate the golden** to 0.7.8 for honesty — the version line is the only change.

## Verification

- `test_manifest_to_findings.py`: **29 passed, 0 failed** (was 27/1).
- `test_render.py`: **314 passed, 0 failed** (unchanged).
- **Future-proofing proven:** temporarily bumped `plugin.json` to `0.9.99` with the golden left at 0.7.8 → test still passes (masked compare ignores the version; the stamping assertion confirms the converter wrote 0.9.99). Restored.

## Files
- `.github/workflows/ci.yml`, `.github/workflows/release.yml` — run the manifest test
- `tests/render/test_manifest_to_findings.py` — mask version in golden compare + assert stamping
- `tests/fixtures/helperbot.from_manifest.json` — regenerated (0.7.5 → 0.7.8)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,5 +23,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Render pipeline tests (schema + render + golden files)
         run: python3 tests/render/test_render.py
+      - name: Manifest→findings converter tests (Step 9.9 → Step 10)
+        run: python3 tests/render/test_manifest_to_findings.py
       - name: Package the distributable (sanity)
         run: bash build.sh
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,9 @@ jobs:
             exit 1
           fi
       - name: Run the test suite
-        run: python3 tests/render/test_render.py
+        run: |
+          python3 tests/render/test_render.py
+          python3 tests/render/test_manifest_to_findings.py
       - name: Install build-only docs dependency
         # Lets build.sh rebuild guide/ and enforce the docs freshness backstop.
         # Build-only; never shipped in the plugin (which stays stdlib-only).
--- a/PRAXEN_SPEC.md
+++ b/PRAXEN_SPEC.md
@@ -59,7 +59,7 @@ Reports must never contain the literal value of a secret, credential, token, pas
 
 ### 2.4 Least privilege
 
-Praxen operates read-only on the agent's workspace. It does not take action on behalf of the agent. It does not modify the agent's code, skill files, configuration, or dependencies. It writes output only to its own `./reports/` directory in the current working directory.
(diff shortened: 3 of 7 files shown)


virtualsteve-exa:
@emmanuelgjr — thanks for the catch on the gated/stale manifest test. This fixes both halves you flagged: the test is now run in CI (`ci.yml` + `release.yml`), and the golden is decoupled from the live version so it won't silently go stale on the next bump. Would appreciate your review when you have a moment.

virtualsteve-exa:
Folded your second round of feedback into this branch (same theme — accuracy of the pipeline's guarantees), @emmanuelgjr:

**Determinism.** You're right that "byte-identical every time" is a property of *rendering*, not findings. The spec's `§9` had one loose line — "every report looks identical, byte-for-byte, regardless of which model produced the findings" — that conflated render-determinism with whole-report identity. Now scoped: the renderer is byte-deterministic (same findings JSON → byte-identical HTML/TXT), while the findings are LLM judgment and vary run to run (exactly why the baselines split sequential vs claude48). The user-facing `understanding-variability.md` already framed this correctly; this just fixes the spec.

**Read-only.** Also right that `§2.4`'s "operates read-only on the agent's workspace" reads like a sandbox when it's a behavioral contract of the skill's instructions — the skill grants Bash and Write, so it's convention, not isolation. Reworded to say so (and to note it's the same declared-vs-enforced distinction Praxen flags when it scans *other* agents). Mirrored the nuance in `docs/usage.md`.

Language only — no technical change. `guide/usage.html` regenerated in lockstep. Both render test suites still green (314/0 and 29/0).

PR #91, opened 2026-06-12
ci(dco): fast-fail + auto-comment the fix on PRs missing sign-off

virtualsteve-exa:
**Problem:** a PR without DCO sign-off currently just shows a red ✗ that nobody sees, so maintainers end up chasing contributors. Worse, for first-time fork contributors the DCO check (on `pull_request`) is *held pending maintainer approval* — so it doesn't even run until you intervene.

**Fix:** move DCO to `pull_request_target` and make it self-service:
- **Fires instantly** for everyone, including first-time fork contributors (not subject to the first-timer workflow-approval hold).
- On a missing/mismatched `Signed-off-by`, **posts a PR comment** (updated in place — no spam) with the exact one-line fix:
  ```
  git rebase --signoff HEAD~N
  git push --force-with-lease
  ```
- Still **fails the check** as the gate; flips the comment to ✅ once fixed.

**Safety:** the job never checks out or runs PR code — it only reads commit metadata via the API and posts a comment. That's the standard, safe `pull_request_target` pattern; no untrusted code runs with the write token. Bot-exempt + merge-skip + author/committer-match rules are unchanged.

**Note on testing:** `pull_request_target` runs the workflow from the *base* branch, so this new version only takes effect once merged. After it lands on `dev`/`main`, a throwaway unsigned test PR will confirm the fast-fail + comment.

Optional belt-and-suspenders (separate, your call): make **DCO a required status check on `dev`** so an unsigned PR also can't be merged, not just flagged.

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -4,62 +4,105 @@
 # Enforces the Developer Certificate of Origin (see CONTRIBUTING.md): every
 # non-merge, non-bot commit in a pull request must carry a `Signed-off-by:`
 # trailer whose name/email matches the commit's author or committer. GitHub
-# bot accounts (Dependabot, gemini-code-assist, …) are exempt — they are
-# not human contributions and the DCO does not apply to them. No third-party
-# action — just git.
+# bot accounts (Dependabot, gemini-code-assist, …) are exempt.
+#
+# Runs on `pull_request_target` (not `pull_request`) on purpose:
+#   - it fires IMMEDIATELY, even for first-time fork contributors (whose
+#     `pull_request` workflows are otherwise held pending maintainer approval),
+#     so a missing sign-off fails fast instead of waiting to be noticed;
+#   - it has a write token, so on failure it posts a PR comment with the exact
+#     one-line fix — the contributor sees it instantly, no chasing required.
+#
+# Safety: this job NEVER checks out or runs the PR's code. It only reads commit
+# metadata through the API and posts a comment. That is the standard, safe use
+# of `pull_request_target` — no untrusted code runs with the write token.
 name: DCO
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: write
 
 jobs:
   dco:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - name: Check Signed-off-by on every commit, comment the fix on failure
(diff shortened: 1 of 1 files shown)


virtualsteve-exa:
Codex review for PR #91, reviewed at head `4dccd34ee95f9af9df066f959d73c155b9294cca`.

Findings:

- **[P2] `.github/workflows/dco.yml:41` — PR commit listing is capped at 250 commits.** The new workflow relies on `github.rest.pulls.listCommits`, but GitHub documents that this endpoint lists a maximum of 250 commits for a pull request: https://docs.github.com/en/rest/pulls/pulls#list-commits-on-a-pull-request. The old `git rev-list "$BASE_SHA..$HEAD_SHA"` path checked the whole range, so after this change a large PR can have commits beyond the API cap that are never DCO-checked and still pass. Suggested fix: use a complete commit source, or explicitly fail when `pr.commits > commits.length` so the gate does not silently pass a partially checked PR.

- **[P3] `.github/workflows/dco.yml:66` — reusable comment lookup should filter to the bot author.** The workflow finds the reusable DCO comment by marker only. Because `<!-- dco-check -->` is public, any PR commenter can include that marker before the workflow runs; the action will then try to update that comment instead of the bot's own comment. Suggested fix: include the author in the match, e.g. `x.user?.login === 'github-actions[bot]' && x.body?.includes(marker)`.

Verification:

- Confirmed PR #91 is still at `4dccd34ee95f9af9df066f959d73c155b9294cca`, mergeable, and only touches `.github/workflows/dco.yml`.
- Ran `git diff --check origin/dev...origin/pr/91`.
- Parsed the workflow YAML successfully with Python/PyYAML.
- Confirmed the configured `pull-requests: write` permission is sufficient for creating/updating issue/PR comments per GitHub's issue-comments docs.
- `actionlint` is not installed locally, so I did not run it without adding tooling.

virtualsteve-exa:
Thanks for the review — both addressed in `1c1044c`:

- **[P2] 250-commit cap:** the workflow now **fails closed** when `pr.commits > commits.length` (the count the API can return) — it posts a "couldn't verify every commit" comment and fails, rather than silently passing a partially-checked PR. Good catch; the old `git rev-list` range didn't have this limit.
- **[P3] comment hijack:** the reusable-comment lookup now matches the **bot author** (`github-actions[bot]`) in addition to the `<!-- dco-check -->` marker, so a pre-posted marker can't redirect the update.

(Gemini couldn't review this one — it doesn't support workflow YAML — which is why a real review here was valuable.)

PR #92, opened 2026-06-12
TEST (do not merge): unsigned commit — validate DCO fast-fail

virtualsteve-exa:
Throwaway PR to validate #91's DCO workflow against the fix branch as base. Should: fail the DCO check + auto-post the fix comment. Will be closed.


virtualsteve-exa:
Closing — served its purpose: proved that the new `pull_request_target` DCO trigger doesn't activate from a feature-branch base (trigger changes only register once on the default branch). Validation will run after #91 lands on main.

PR #94, opened 2026-06-12
TEST (do not merge): validate live DCO fast-fail

virtualsteve-exa:
Throwaway unsigned PR to confirm the now-live pull_request_target DCO workflow fires + auto-comments. Will be closed.


virtualsteve-exa:
Closing — validation complete. The live pull_request_target DCO workflow fired instantly, failed on the unsigned commit + auto-posted the one-line fix, then flipped the same comment to ✅ once signed. Working as intended.

PR #96, opened 2026-06-12
ci(dco): stay quiet on clean PRs; only flip an existing miss to ✅

virtualsteve-exa:
The DCO check posted a `✅ DCO check passed — thanks!` comment on **every** PR — including ones signed off from the first push. That's noise: the green status check already says it passed.

## Change
- **Success path posts no new comment.** It only updates an *existing* DCO comment (one we'd posted to flag a miss), flipping ❌ → ✅ (`DCO check now passes — thanks for fixing it!`) so a contributor who just signed off sees the loop closed.
- A PR that's clean from the start now gets **zero** comments; a PR that missed then fixed gets **one** comment that ends green.
- **Failure behavior unchanged** — still fails fast and comments the one-line fix.
- Refactor: split the listComments/find into `findDcoComment()`, shared by the upsert (failure) and update-only (success) paths.

## Note on testing
Like the original, this is a `pull_request_target` workflow, so the version that runs is the one on the PR's **base** branch — it can't be exercised from this PR's head. It'll take effect for PRs targeting `dev` once merged here, and for `main` once promoted. YAML + embedded JS validated locally (`node --check`, async-wrapped as github-script runs it).

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -38,15 +38,21 @@ jobs:
             const pr = context.payload.pull_request;
             const marker = '<!-- dco-check -->';
 
-            // Post or update the single bot-authored DCO comment (never spam,
-            // and never hijackable: match the bot author, not just the marker).
-            async function upsertComment(body) {
+            // Find the single bot-authored DCO comment, if one exists (match the
+            // bot author, not just the marker, so it can't be spoofed).
+            async function findDcoComment() {
               const comments = await github.paginate(github.rest.issues.listComments, {
                 owner, repo, issue_number: pr.number, per_page: 100,
               });
-              const existing = comments.find(
+              return comments.find(
                 x => x.user && x.user.login === 'github-actions[bot]' && x.body && x.body.includes(marker)
               );
+            }
+
+            // Post the DCO comment, or update it in place if we've already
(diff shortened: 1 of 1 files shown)


virtualsteve-exa:
Codex review note for PR #96, reviewed at head `664f7a878bf96e950fc80096e049ce4ad6f594fc`.

No findings.

I focused on `.github/workflows/dco.yml`, especially the success/failure comment behavior inside the `pull_request_target` GitHub Script. The refactor keeps the previous hardening intact: the reusable DCO comment is matched by `github-actions[bot]` plus marker, and the >250-commit API cap path still fails closed instead of partially checking a PR.

Verification run:

- `git diff --check origin/dev...origin/pr/96`
- Parsed `.github/workflows/dco.yml` successfully with PyYAML.
- Extracted the embedded `github-script` body and checked it with `node --check` inside an async wrapper.
- Mock-executed the actual embedded script for the important paths:
  - clean PR with no existing DCO comment -> no comment created
  - clean PR with an existing DCO miss comment -> updates that bot comment to ✅
  - unsigned commit -> creates the failure comment and calls `core.setFailed`
  - 251 reported commits / 250 visible commits -> creates the warning comment and fails closed

`actionlint` is not installed locally, so I did not run that without adding tooling.

Issue #101, opened 2026-06-15
1.0 Release Review — audit checklist & gate tracking


virtualsteve-exa:
**Audit round 1 complete → MEDIUM fixes landed.** All pre-tag hard blockers that *can* be cleared before tagging are cleared; what remains is tag-time actions + a few decisions. Findings + the 1.0 stability-contract draft are in the comments.

**Legend:** 🔴 hard blocker · 🟡 strong · ⚪ polish  ·  status: ✅ done · ⏳ tag-time only · ◻️ decision owed · ⬜ open (not blocking)

---

## 0. The 1.0 stability contract (define first)
- [x] 🟡 ✅ Stability contract adopted as `STABILITY.md`.
- [x] 🟡 ✅ Semver / compatibility policy published (in `STABILITY.md`, linked from README).

## 1. Coding-agent-skill
- [x] 🔴 ✅ Plugin manifest schema valid; `source` not bare `"."` — now **machine-tested in CI** (`test_plugin_manifests.py`, M7).
- [ ] 🔴 ⏳ Post-tag install propagation on both Claude Code + Codex.
- [x] 🟡 ✅ Install / uninstall paths verified (one-liner + uninstall flow exercised this cycle).
- [x] 🟡 ✅ Model-support statement + cross-model comparability caveat (M4).
- [x] ⚪ ✅ Step 10 synthesis burst eliminated (#32 closed: `manifest_to_findings.py`); large-workspace mitigations in place.
- [x] 🟡 ✅ Render pipeline deterministic; golden + baseline re-render green on Py 3.9/3.12/3.13.

## 2. Security-tool posture
- [x] 🔴 ✅ HTML-report XSS — escaped + **regression test added** (mutation-proven).
- [ ] 🟡 ⬜ Prompt-injection resistance (hostile-target fixture) — not yet tested.
- [x] 🔴 ✅ Never-reprint-secrets — code backstop added (warn, M2) + test. (warn-only by decision)
- [x] 🟡 ✅ "Nothing phones home" true for the shipped skill (stdlib-only).
- [x] 🟡 ✅ Read-only / external-observer posture.
- [ ] 🟡 ⬜ Responsible-disclosure hygiene in shipped examples/baselines — not re-reviewed.
- [x] 🔴 ✅ Supply chain: deps pinned; #80 resolved; **Mermaid exact-pin + SRI** (M3).
[...]

Issue #100, opened 2026-06-15
RFE: docs consistency — medium-priority editorial items deferred from #99


virtualsteve-exa:
## Summary

Follow-up to the pre-1.0 docs editorial audit. The **BIG** consistency items shipped in #99 (sentence-case headings, divider cleanup, em-dash list style, blockquote-callout system, four-input-shapes + official MCP-title spec sync). This RFE parks the **MEDIUM** items — real consistency gaps, but not blockers — so we can pick them up when convenient. None is urgent; all are low-risk editorial.

The conventions for several of these are already codified in the new [`docs/README.md`](../blob/main/docs/README.md) style guide; the work here is mostly **applying those rules to existing prose** plus a few small structural fixes.

## Medium items

- [ ] **Add "Next steps" footers to the two pages that dead-end.** Nine docs have a `## Next steps` footer; `usage.md` (the longest, most procedural page) and `RAISE.md` (a heavy conceptual page) don't. Add them, linking to the natural next reads.
- [ ] **Standardize the output-file placeholder.** Docs mix `<agent>` (`index.md`, `interpreting-reports.md`) and `<agent-slug>` (`usage.md`, `PRAXEN_SPEC.md`). Pick **`<agent-slug>`** (it's what the implementation actually uses) everywhere. *Touches `PRAXEN_SPEC.md` — keep it synced.*
- [ ] **Decide on timestamp-format precision.** Convenience docs use generic `<timestamp>` / `<date>`; the spec's Outputs table uses explicit `<YYYY-MM-DD-HHMMSS>` (HTML/TXT) and `<YYYY-MM-DD>` (JSON). Either spell out the formats in the convenience docs, or add a one-line note that JSON is per-day while HTML/TXT are per-run.
- [ ] **Point-of-view pass on procedural pages.** Action pages mix direct "you" with passive/descriptive voice. Standardize on second-person "you" (the operator) for procedural pages, leaving conceptual pages (`abv.md`, `RAISE.md`, `owasp.md`) descriptive. *(The `docs/README.md` style guide already states this rule — this applies it.)*
- [ ] **Reconcile the quickstart timing claim.** `quickstart.md` promises "5 minutes" while `usage.md` says scans "typically take 8–15 minutes." Make clear the quickstart figure is the tiny bundled `finbot` example, not a general scan time. While there, fix the numeral: `quickstart.md`'s H1 says "**5** minutes" but four other docs say "**five** minutes" — the style guide says spell out small numbers.

[...]

PR #88, opened 2026-06-11
docs(codex): skill-folder install only; drop the marketplace one-liner

virtualsteve-exa:
Codex validation (#81) confirmed the public `codex plugin marketplace add` one-liner can't expose the repo-root plugin, and every no-copy fix is a symlink-trick or a maintained duplicate. The verified, simple Codex install is the **skill-folder path** — so make that the whole story and stop chasing the one-liner.

- Remove `.agents/plugins/marketplace.json`; drop `.agents` from the zip allowlist.
- `installation.md`: Codex (Option B) is now skill-folder-only (removed the experimental-marketplace note).
- README: tidy the install-link line.
- Keep `.codex-plugin/plugin.json` (standard Codex plugin descriptor; 4-way version guard intact).
- `PRAXEN_SPEC.md` §10: firm up the model claim — validated on GPT-5-class.

No version change. The skill-folder Codex install was smoke-tested end-to-end on Codex (Step 9.9 → both Python scripts → reports, under workspace-write).

Issue #69, opened 2026-06-01
RFE: replace the `sweep` baseline target — dead upstream + redundant archetype


virtualsteve-exa:
## Summary

Retire `sweep` from the baseline suite and replace it with a target that **adds an under-covered archetype** (not another coding agent), on a live/maintained upstream.

This is the *second* prune in the autonomous-coding cluster. The first — `devika` → `fetchai-uagents` — is already staged (`tests/baselines/v0.7.9-pre/`). `sweep` is the next one; filing it separately so we can loop back.

## Why `sweep` is low-impact to remove

**1. Dead upstream.** `sweepai/sweep` — 7.7k★ but last pushed **2025-09-18** (~8.5 months stale); sweepai pivoted off the product. A dormant repo tests a frozen threat surface that no longer evolves.

**2. Most over-represented archetype.** Autonomous coding agents are **5 of 12** baseline targets (`sweep`, `devika`, `aider`, `openhands`, `deepagents-cli`). Three of the five — `aider` (45.6k★), `openhands` (75.5k★), `deepagents` (23.6k★) — are on big, *actively-developed* repos and keep the archetype well covered.

**3. Owns zero unique coverage.** Every OWASP category in `sweep`'s findings (ASI01/02/03/05, LLM01/02/05/06) appears in at least one other target. Removing it loses **no** category coverage. (Checked against the v0.7.7-claude48 set.)

**4. Maturity is mid-band and covered.** Weighted RAISE 1.30 (Ad hoc) — well within the range the rest of the suite already spans.

## The ask

Replace `sweep` with a live target that **broadens** the suite rather than deepening the coding-agent cluster — e.g. an under-represented archetype such as a browser-use/computer-use agent, a voice/realtime agent, a RAG/retrieval agent, or a multi-agent orchestration framework. Selection should favor an **actively-maintained, popular upstream** (the repo-health signal that flagged `sweep` in the first place).

Keep `finbot` (CTF anchor) and `helperbot` (uniquely owns LLM07) — they earn their slots despite low stars; they're purpose-built security demos, not popularity contests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

virtualsteve-exa:
Executed in **v1.0.2** (2026-07-11): sweep retired from the roster (dead upstream + redundant autonomous-coding archetype), replaced in spirit by the CraftBot/uAgents/Agentforce additions. Closing as done — post-1.1 hygiene sweep (PR #187).

PR #103, opened 2026-06-15
prep(1.0): security backstops, CI manifest gate, stability contract, docs polish

virtualsteve-exa:
Pre-1.0 hardening batch — **no version bump** (still `0.8.1`). Rolls up the [#101](https://github.com/open-agent-ai-security/praxen/issues/101) audit's MEDIUM fixes, the deferred docs polish ([#100](https://github.com/open-agent-ai-security/praxen/issues/100)), and [#50](https://github.com/open-agent-ai-security/praxen/issues/50). Nothing here changes the scan engine's scoring logic.

## Security
- **XSS regression test** (`test_render.py` §4f) — injects `<script>`, `onerror`, attribute-breakout, `javascript:`, and tag-smuggling into every untrusted field, renders, and asserts none execute. Mutation-proven: it fails if `render.py`'s escaping is removed.
- **Secret backstop** (`render.py`) — warns on **stderr** (location + pattern, never the value) when an evidence snippet looks like a credential. Output is unchanged, so the render stays byte-deterministic (golden unaffected). Covered by tests.

## CI / contract
- **`tests/render/test_plugin_manifests.py`** wired into `ci.yml` — catches the bare-`"."` marketplace `source` (the v0.6.1 install-breaker) and version drift across the three manifests + `PRAXEN_SPEC.md`, on every PR. Mutation-tested.
- **`findings.schema.json`** `schema_version` → `^2\.[0-9]+$` (was `const "2.0"`), matching the minor-tolerant validator.
- **`STABILITY.md`** — the 1.0 stability contract (schema, RAISE categories + weights, skill/output interface, remit template) + a one-paragraph semver policy. Linked from README.

## Supply chain
- **Mermaid pinned to `11.15.0` + SRI integrity hash** (was a `@11` range, no integrity) — fail-closed; MIT attribution added to `NOTICE`. Diagrams verified rendering over HTTP. `guide/` regenerated.

## Docs
- **Model-tier statement + cross-model score-comparability caveat** (README, `understanding-variability.md`).
- **`SECURITY.md`** supported-versions de-staled; **`CONTRIBUTING.md`** gains a "Releasing and rolling back" runbook.
- **#100 consistency polish:** `RAISE.md` Next-steps footer; `<agent-slug>` placeholder unified; timestamp-precision note; quickstart timing + numeral; "findings JSON" short-name; draft-manifest clarification. *(POV and "OWASP …Project" left unchanged — both are correct as-is; see #100 close comment.)*
- **CHANGELOG** `[Unreleased]` section + a correction to the **0.8.1** entry recording the #88 `.agents/` removal it had omitted.

## Fixes
- `owasp_coverage.py` — harden the `as_uri()` fallback against a path-like `str` (**Closes #50**).

## Verification
`render 329 / manifests 17 / converter 29`, all green · 0 broken doc links/anchors · `docs_build.py` idempotent · DCO signed.

## Out of scope (1.0 tag-time / decisions)
Full 12-target suite + the `1.0.0` bump + post-tag install checks are tag-time. Open decisions tracked in #101: CI-review replacement (Gemini sunset), baseline-target refresh (#69/#70/#64).


virtualsteve-exa:
Reviewed PR #103, with a hard look at Issue #100 as requested.

Findings:

- **[P3] Issue #100 is not fully closed.** `PRAXEN_SPEC.md` still shows `<agent>` / `<ts>` in the architecture diagram while the outputs table below uses `<agent-slug>` and explicit timestamp formats. Also `docs/quickstart.md` and `docs/usage.md` still tell readers to expect "three files" without listing the draft checkpoint. That leaves the exact consistency gaps #100 was tracking. Relevant spots: `PRAXEN_SPEC.md:107`, `PRAXEN_SPEC.md:116`, `PRAXEN_SPEC.md:121`, `docs/quickstart.md:59`, `docs/usage.md:93`.

- **[P3] Mermaid authoring docs are stale after the SRI change.** `docs_build.py` now intentionally uses the classic UMD build because ESM cannot carry integrity, but `docs/README.md` still says the runtime is a remote ES module blocked from `file://`. The preview/gotcha guidance should be updated to match the new implementation. Relevant spots: `docs_build.py:68`, `docs/README.md:114`.

Verification performed locally on PR head `306b146`:
- `python3 tests/render/test_render.py` -> 329 passed, 0 failed
- `python3 tests/render/test_manifest_to_findings.py` -> 29 passed, 0 failed
- `python3 tests/render/test_plugin_manifests.py` -> 17 passed, 0 failed
- `python3 docs_build.py` clean and idempotent
- `./build.sh`
- `python3 -X pycache_prefix=/tmp/praxen-pr103-pycache -m py_compile docs_build.py skills/behavior-verifier/render.py tests/render/test_plugin_manifests.py tests/baselines/owasp_coverage.py`
- `git diff --check origin/dev...HEAD`
- generated `guide/*.html` internal link / anchor check clean
- Markdown docs link target check clean
- Mermaid `11.15.0` SRI hash verified against the jsDelivr artifact

No other blocking issues found in the renderer/security backstop, manifest test, schema-version tolerance, package build, or generated guide output.

virtualsteve-exa:
Thanks both — all review comments addressed in `fa9c7fd`.

**@gemini-code-assist (medium) — `test_plugin_manifests.py` plugins loop:** fixed. The loop now uses `for e in (plugins if isinstance(plugins, list) else []):`, matching the defensive pattern already used for `entry`. A malformed non-list `plugins` can no longer crash the test.

**[P3] #100 not fully closed — `PRAXEN_SPEC.md` diagram + "three files":** good catch, that was a real miss (I'd synced `docs/` but not the spec).
- The §3 architecture diagram now uses `<agent-slug>` / `<timestamp>` throughout (dropped the redundant `./reports/` prefix so the longer placeholders fit the box without re-aligning it; the outputs table right below already states the directory and uses bare filenames). `grep '<agent>' PRAXEN_SPEC.md` → 0.
- `quickstart.md` and `usage.md` now mention the working `<agent-slug>-draft-<timestamp>.md` checkpoint alongside the "three files," so the 4th file isn't a surprise (matching the note already in `interpreting-reports.md`).

**[P3] Mermaid authoring docs stale after the SRI change:** fixed. `docs/README.md` no longer says the runtime is a remote ES module — it now correctly describes the classic UMD CDN script loaded with SRI + `crossorigin`, and frames the `file://` fallback around the integrity-checked cross-origin fetch (browser-dependent) rather than the old ESM-import reason.

Re-verified: render 329 / manifests 17 / converter 29, 0 broken doc links, `docs_build.py` idempotent. (build.sh's guide-freshness check passes once the regenerated `guide/` is committed, which it now is.)

virtualsteve-exa:
Re-review on `fa9c7fd` after the latest fixes. The earlier Issue #100/docs findings look addressed: placeholder names, draft-checkpoint wording, timestamp precision, and the Mermaid/SRI docs now line up. I found two remaining issues:

- [P2] The schema-version contract is ahead of the validators. `STABILITY.md` says 2.x minors may add optional fields and enum values, and `findings.schema.json` now accepts any `2.x`. But the published schema is still closed with `additionalProperties: false`, and the runtime still rejects unknown enum values via `schema.py`'s fixed enum helper. This advertises future `2.1` compatibility that the tooling cannot actually honor. Either keep this pinned to exact `2.0` for now, or make schema/runtime truly minor-tolerant.

- [P2] The secret “backstop” warns but still writes leaked values into HTML reports. `render.py` scans evidence snippets, emits stderr, then renders as-is; `test_render.py` codifies that behavior. I reproduced this with the FinBot fixture plus an AWS-shaped key in evidence: `--out-html` exits 0, stderr omits the value, but the generated HTML contains it. That conflicts with the spec's “Never reprint secrets” rule. This should fail before writing or redact before rendering.

Checks run locally: `python3 tests/render/test_render.py` (329 passed), `python3 tests/render/test_manifest_to_findings.py` (29 passed), `python3 tests/render/test_plugin_manifests.py` (17 passed), `python3 docs_build.py`, `python3 -X pycache_prefix=/tmp/praxen-pr103-pycache -m py_compile docs_build.py skills/behavior-verifier/render.py tests/render/test_plugin_manifests.py tests/baselines/owasp_coverage.py`, `git diff --check origin/dev...HEAD`, generated guide/docs link+anchor check, and `bash build.sh`. Worktree was clean after checks.

virtualsteve-exa:
Both P2s addressed in `c4390ab`.

**[P2] Secret backstop now redacts, not just warns.** You were right — warn-only left the value in the shareable report, which doesn't satisfy Never-Reprint-Secrets. `render.py`'s `_scrub_secrets` now **replaces** a credential-shaped evidence snippet with `[REDACTED — <pattern>]` before rendering, so the value never reaches the HTML/TXT. It still warns on stderr (location + pattern, never the value), operates on the in-memory copy only (on-disk JSON untouched), and stays deterministic. Verified end-to-end with your repro (FinBot + an AWS-shaped key): `--out-html` exits 0, **the value is absent from the HTML** and `[REDACTED — AWS access key id]` appears in its place. The test now asserts the value is absent from HTML/TXT and that `[REDACTED]` is present. *(This supersedes the earlier warn-only call — redaction is strictly stronger: it warns **and** prevents the leak.)*

**[P2] Schema-version contract made honest.** Took your first option — keep it pinned to exact `2.0` rather than loosen the strict validator:
- `findings.schema.json` `schema_version` → `const "2.0"` with an honest description (reverts the transient "accepts any 2.x" from earlier in this PR).
- `schema.py` validates `schema_version` as **exact** `2.0` (was major-tolerant — which was the source of the hollow promise: it accepted a `2.1` version string then rejected its content).
- `STABILITY.md` reframes additive-minor compatibility as a **policy for future minors** (published schema + validator updated in lockstep when one ships) plus consumer-side parsing guidance — not a claim that today's strict tooling reads a future minor.
- No behavior change vs `0.8.1` (all committed baselines are `2.0`); it documents and tightens the existing contract rather than loosening it.

Re-verified: render **331** / manifests 17 / converter 29, goldens byte-identical, `py_compile` clean. Both were good catches — the secret one especially; thanks for the repro.

virtualsteve-exa:
Re-review on `c4390ab`: the schema-version P2 looks fixed, and the AWS-shaped-key repro is fixed. I found one remaining issue:

- [P2] Private-key redaction only removes the header, leaving the key body in the HTML report. In `render.py`, the `private key block` regex matches only `-----BEGIN ... PRIVATE KEY-----`; `_scrub_secrets` then substitutes only that header. I reproduced with a multi-line private key snippet: render exits 0, warns that it redacted a private key block, removes the BEGIN line, but the base64 key body remains in the generated HTML. That still violates the Never-Reprint-Secrets rule. The fix should redact the whole PEM block, e.g. BEGIN through matching END with a DOTALL regex, and add a regression test for that full-block case.

Checks run locally: `python3 tests/render/test_render.py` (331 passed), `python3 tests/render/test_manifest_to_findings.py` (29 passed), `python3 tests/render/test_plugin_manifests.py` (17 passed), `python3 docs_build.py`, `python3 -X pycache_prefix=/tmp/praxen-pr103-pycache -m py_compile docs_build.py skills/behavior-verifier/render.py skills/behavior-verifier/schema.py tests/render/test_plugin_manifests.py tests/baselines/owasp_coverage.py`, `git diff --check origin/dev...HEAD`, guide/docs link+anchor check, and `bash build.sh`. Worktree was clean after checks.

virtualsteve-exa:
Fixed in `097ad55` — and thanks, that was a real leak in my own redaction.

**[P2] Full PEM private-key block now redacted entirely.** The `private key block` pattern matched only the `-----BEGIN…-----` header, so the base64 body survived. It now spans the **whole block** — `BEGIN` through matching `END` with `re.DOTALL` — exactly as you suggested. I also added a `|.*` fallback so a **truncated** block (BEGIN + body, no END line) redacts to the end of the snippet rather than leaking the body.

Verified end-to-end on both shapes (full block and truncated): the key body is **absent from the rendered HTML**, with `[REDACTED — private key block]` in its place. Added the full-block regression test you asked for (render is now **332**).

Re-verified: render 332 / manifests 17 / converter 29.

virtualsteve-exa:
Final re-review on `097ad55`: no findings.

I re-ran the private-key repros against this head: both full PEM blocks and truncated PEM blocks now redact cleanly, with no header/body/footer leakage into HTML or TXT. The previous schema-version and AWS-shaped-key findings also remain fixed.

Checks run locally: `python3 tests/render/test_render.py` (332 passed), `python3 tests/render/test_manifest_to_findings.py` (29 passed), `python3 tests/render/test_plugin_manifests.py` (17 passed), `python3 docs_build.py`, `python3 -X pycache_prefix=/tmp/praxen-pr103-pycache -m py_compile docs_build.py skills/behavior-verifier/render.py skills/behavior-verifier/schema.py tests/render/test_plugin_manifests.py tests/baselines/owasp_coverage.py`, `git diff --check origin/dev...HEAD`, guide/docs link+anchor check, and `bash build.sh`. Worktree was clean after checks and `HEAD` matched `origin/pr/103`.

Issue #105, opened 2026-06-15
1.0 Codex install + first scan UX findings


virtualsteve-exa:
## Summary

Codex support is functionally working on the 1.0 RC path, including discovery of the `1.0.0-rc.1` skill, execution through the Praxen manifest/converter/renderer pipeline, and a real FinBot single-target scan. The main 1.0 polish opportunity is documentation/operator UX: the current docs are mostly accurate, but a first-time operator can easily confuse `examples/` demo outputs with scan inputs, and the Codex install/run path needs a more explicit happy path.

## What I verified

Environment:

- Praxen branch: `rc/1.0.0-codex-check`
- Praxen commit: `28dcd154bf692ac961a0fde3691f636c38167d38`
- Praxen manifest version: `1.0.0-rc.1`
- Codex CLI: `OpenAI Codex v0.140.0-alpha.2`
- Codex model: `gpt-5.5`
- Codex sandbox: `workspace-write`

Prerelease/version check:

- `.codex-plugin/plugin.json` contained `"version": "1.0.0-rc.1"`.
- `build.sh` accepted the prerelease SemVer and produced `dist/praxen-1.0.0-rc.1.zip`.
- Codex loaded the skill without any version-related error.
- Skill surfaced as `praxen:behavior-verifier`.
- `$praxen:behavior-verifier` resolved to `skills/behavior-verifier/SKILL.md`.

Real single-target scan:

- Source cloned from `https://github.com/OWASP-ASI/finbot-ctf-demo`.
- Remit copied from `tests/remits/finbot.md` to `WORKER_REMIT.md`.
- Ran Codex from a fresh scan directory against `./finbot-ctf-demo`.
- The scan completed through:
  - draft manifest
  - `manifest_to_findings.py`
  - `render.py`
  - schema validation
  - HTML/TXT rendering

Generated outputs:

- `reports/finbot-draft-2026-06-15-210423.md`
- `reports/finbot-findings-2026-06-15.json`
- `reports/finbot-analysis-2026-06-15-210423.html`
- `reports/finbot-analysis-2026-06-15-210423.txt`

Host-side validation:

- JSON had `praxen_version: "1.0.0-rc.1"`.
- JSON had `schema_version: "2.0"`.
- Re-render from JSON passed independently: `11 findings, 0 schema errors`.
- FinBot scan result: `7 Critical / 2 High / 2 Medium`, weighted RAISE `1.00 / 5.0`.

## What worked well

[...]

Issue #102, opened 2026-06-15
RFE: public Codex install path — `codex plugin marketplace add` / `.agents/plugins/marketplace.json`


virtualsteve-exa:
Split out from #81 (Codex support), which delivered the core: `.codex-plugin/plugin.json` (version-synced and validated by `build.sh`), the shared `skills/behavior-verifier/` engine loaded by both platforms, and Codex install docs (`docs/installation.md` Option B — the `.agents/skills/` symlink + `$praxen:behavior-verifier` invocation).

**Remaining:** a **public, one-line Codex install path** equivalent to the Claude `claude plugin marketplace add …`. Codex can load skills from a local `.agents/skills` location or a Codex plugin bundle; what's missing is a published marketplace/bundle manifest (e.g. `.agents/plugins/marketplace.json`) so Codex users get a copy-paste install instead of hand-wiring the symlink.

When this lands:
- If `.agents/` becomes a shipped path, update `build.sh`'s INCLUDE list (flagged in the 1.0 release-engineering audit, #101) or it will be silently omitted from the zip.
- Add the Codex one-liner to `README.md` / `docs/installation.md` alongside the Claude one.

**Not a 1.0 blocker** — the documented unzip + `.agents/skills/` path works today.

virtualsteve-exa:
**Decision: holding this for 1.0.** The documented symlink path stays the first-class Codex install; the public marketplace path waits. Logging the new data from the #105 RC verification below since it bears on this RFE.

**The symlink/unzip path is now empirically verified end-to-end** (previously assumed-working). On the `rc/1.0.0-codex-check` tree — Codex CLI **v0.140.0-alpha.2**, model **gpt-5.5**, `1.0.0-rc.1` manifest — the skill was discovered as `praxen:behavior-verifier` via `.agents/skills/` and ran a full real FinBot scan through the pipeline. The team used the symlink; no `codex plugin marketplace` command was exercised (so whether that subcommand even exists yet in the current Codex CLI is still the open dependency for this RFE).

**New friction data that strengthens the case for an eventual packaged install:**
[...]

PR #107, opened 2026-06-15
docs(#105): examples-vs-scan-targets clarity + Codex operator UX

virtualsteve-exa:
Acts on the Codex test team's 1.0 RC verification report (#105).

**Context — the RC test passed.** Codex (CLI v0.140.0-alpha.2, gpt-5.5, workspace-write) loaded the `1.0.0-rc.1` skill with no version error and ran a real FinBot single-target scan through the full pipeline, preserving every Critical theme from the Claude baseline. No `0.9.0` fallback needed. The report's findings were all documentation / operator-UX, addressed here.

### What changed
- **`examples/` are outputs, not scan targets** (the #1 first-run footgun). Loud callout added to the README *Examples* section, `examples/README.md` top block, and the `tests/README.md` single-target section; `docs/quickstart.md` §1 reframed so the **remit** is the one input taken from `examples/finbot/` and the **source** is cloned from upstream. Fixed the misleading "scan the bundled `finbot` example" phrasing in README + `installation.md` (Quickstart actually clones the upstream source).
- **Codex install modes** documented: B1 repo-local / B2 user-wide symlink as the current first-class path, with the **working-directory scope gotcha** spelled out (a repo-local `.agents/skills` link isn't visible from a separate scan dir). The not-yet-built public Codex marketplace path links to **RFE #102**.
- **Copy-paste FinBot Codex smoke recipe** added to *Verifying the install*, using a gitignored `local/` scan dir; documented `codex exec --skip-git-repo-check` / avoid-`/tmp` guidance.
- **"identical analyses" softened** → same engine/schema and report format; themes match but counts/grouping/RAISE can vary by model and context (links to Understanding Run-to-Run Variability).
- **Troubleshooting note** for benign Codex loader/telemetry warnings (incl. the colon-id metric-tag warning — a Codex-side constraint, not a Praxen error).
- **Removed the hardcoded `Version 0.8.1`** from the README masthead; the auto-updating *Latest release* badge is the source of truth (no drift during an RC).

### Mapping to #105
| #105 item | Status |
|---|---|
| 1 examples-are-outputs callout | ✅ done (5 spots) |
| 2 copy-paste Codex recipe | ✅ done |
| 3 Codex install modes | ✅ symlink path documented; marketplace path → RFE #102 |
| 4 `--skip-git-repo-check` / avoid `/tmp` | ✅ done |
| 5 soften "identical analyses" | ✅ done |
| 6 benign Codex warnings note | ✅ doc note; telemetry tag is Codex-side (to route to that team) |
| 7 README version drift | ✅ hardcoded string removed |

### Tests
- Regenerated `guide/installation.html` + `guide/quickstart.html` from the markdown (`docs_build.py`).
- `tests/render/test_render.py` → **332/0**; `tests/render/test_plugin_manifests.py` → **17/0**.

Docs-only; no skill/render/schema behavior change.


virtualsteve-exa:
Review findings from the Codex pass:

- [P2] `docs/installation.md:105` uses `tests/remits/finbot.md` in a smoke test that also claims to work from an unzipped release. The release zip intentionally excludes `tests/` (`build.sh:97`), so release users will fail at the `cp`. Suggest using `examples/finbot/WORKER_REMIT.md`, or splitting the checkout-only and release-zip instructions.

- [P2] `docs/installation.md:97-99` sets `PRAXEN_ROOT`, but then creates `local/finbot_scan` relative to the caller current directory. The "repo-local + gitignored, no --skip-git-repo-check needed" promise is only true if the user is already in `$PRAXEN_ROOT`. Suggest adding `cd "$PRAXEN_ROOT"` after the variable assignment, or using absolute paths for the mkdir/cd.

- [P3] `docs/quickstart.md:12` still says installation is "one command from the marketplace," and `docs/quickstart.md:46` is Claude-only. Since Codex is now first-class but uses the link flow for now, I would neutralize that sentence and add a tiny Codex invocation variant.

- [P3] `tests/README.md:70` still frames single-target scans as Claude Code only. Since this is the doc we pointed platform testers at, add a Codex note or cross-link to the Codex smoke-test command.

Validation: `git diff --check` passed. I also confirmed with `codex debug prompt-input` that a repo-local `.agents/skills` link at the repo root exposes `praxen:behavior-verifier` from both the repo root and a nested `local/...` scan directory, so that part of the new flow looks sound.

virtualsteve-exa:
Both review findings addressed in `8691bc5` — both valid:

1. **`SKILL.md`-by-absolute-path in the Codex working-dir note** — correct, that's a Claude Code pattern; Codex discovers via `.agents/skills/`. Dropped it from the note; kept the two valid `.agents/skills/` fallbacks (scan-dir-local, user-wide).
2. **Clone-dir → report-slug mismatch** — correct, `git clone …/finbot-ctf-demo` yields a `finbot-ctf-demo/` workspace and the slug derives from the dir name, so outputs could be `finbot-ctf-demo-*` rather than the documented `finbot-*`. Now `git clone … finbot` and scan `./finbot`, with a one-line note on *why* the dir name matters — deterministic filenames either way the model derives the slug.

`guide/installation.html` regenerated from the markdown; render smoke 332/0.

virtualsteve-exa:
Second-pass review after `8691bc5`:

- [P2] `docs/installation.md:106` still copies `tests/remits/finbot.md` in a smoke test that says it works from "your checkout or unzipped release." The release zip excludes `tests/` (`build.sh:97`), so zip users hit a dead command. Suggest using `examples/finbot/WORKER_REMIT.md` for this smoke test, or splitting checkout-only vs release-zip commands.

- [P2] `docs/installation.md:97-99` still sets `PRAXEN_ROOT` but creates `local/finbot_scan` relative to whatever directory the user is currently in. The "repo-local + gitignored / no --skip-git-repo-check" behavior only holds if they already `cd` into `$PRAXEN_ROOT`. Suggest adding `cd "$PRAXEN_ROOT"` or using absolute paths.

- [P3] `docs/installation.md:47` says repo-local skill links will not be found from a separate scan directory, but the smoke test later relies on a nested `local/finbot_scan` seeing the repo-root link. My Codex debug check showed that nested repo-local scan dirs do see the root `.agents/skills` link, so I would clarify this as "outside the repo tree / outside an ancestor with `.agents/skills`."

- [P3] `docs/quickstart.md:12` and `docs/quickstart.md:46` remain Claude-shaped: "one command from the marketplace" plus "From a Claude Code session." For first-class Codex support, make the install sentence platform-neutral and add a small Codex invocation variant.

- [P3] `tests/README.md:70` still says to open a Claude Code session for single-target scans. Since this is the doc we send platform testers to, it should mention Codex or link back to the Codex smoke-test flow.

Validation: `git diff --check` passed, and `python3 docs_build.py` regenerated cleanly with no working-tree changes.

virtualsteve-exa:
Second-pass review addressed in `edf1c65` — all five, all valid:

- **[P2] zip-unsafe remit** (`installation.md:106`) — right, `tests/` is excluded from the release zip, so `cp tests/remits/finbot.md` is a dead command for zip users. Switched to **`examples/finbot/WORKER_REMIT.md`**, which ships in the zip → the recipe now works from both a checkout and an unzipped release.
- **[P2] `local/` relative to caller cwd** (`installation.md:97-99`) — added **`cd "$PRAXEN_ROOT"`** before the `mkdir`, so `local/finbot_scan` lands under the repo (gitignored + trusted in a checkout). Also called out that an **unzipped release** isn't a git repo, so that case still needs `--skip-git-repo-check`.
- **[P3] gotcha note too absolute** (`installation.md:47`) — corrected per your `codex debug` finding: Codex walks **up** ancestors for `.agents/skills/`, so a nested `local/<t>_scan` *does* see a repo-root link. Reframed the gap as scan dirs **outside the repo tree**, not merely "separate" ones. Thanks for verifying that empirically — it also confirms the smoke recipe's nested scan dir is sound with a single repo-root link.
- **[P3] quickstart Claude-shaped** (`quickstart.md:12,46`) — neutralized the "one command from the marketplace" line and added a Codex invocation variant in §3.
- **[P3] tests/README Claude-only** (`tests/README.md:70`) — added a Codex note cross-linking Installation Option B + the Codex smoke test.

Guide regenerated; render 332/0; `git diff --check` clean.

virtualsteve-exa:
Higher-level Codex docs simplification note for 1.0:

The latest fixes make the details more accurate, but the Codex install path is still carrying too much platform machinery in the happy path. Compared with Claude Code, the Codex section is doing four jobs at once:

1. Installing/discovering the skill via `.agents/skills`
2. Explaining repo-local vs user-wide discovery behavior
3. Explaining Codex trusted-git-directory and workspace-write sandbox behavior
4. Running a real Praxen smoke test with remit + cloned upstream source

That is why the section keeps accumulating caveats around `local/`, ancestor `.agents/skills`, release-vs-checkout paths, `--skip-git-repo-check`, and which remit path ships in the zip. The facts are mostly right now, but the information architecture still makes Codex feel much harder than it is.

Suggested 1.0 simplification:

- Make the default Codex install path **user-wide skill link**, not repo-local:

```bash
cd /path/to/praxen
mkdir -p "$HOME/.agents/skills"
ln -sfn "$PWD/skills/behavior-verifier" "$HOME/.agents/skills/behavior-verifier"
```

- Keep the smoke test separate from install/discovery.
- Split the smoke test into two explicit variants instead of one block plus caveats:
  - **Git checkout:** use `local/finbot_scan`, no `--skip-git-repo-check`
  - **Unzipped release:** use the same scan shape, but include `--skip-git-repo-check` directly in the command
- Move the repo-local discovery behavior, ancestor lookup, sandbox writes, and trusted-directory notes into a short Codex troubleshooting/details subsection.

In other words: give Codex a small golden path first, then explain the edge mechanics after. Right now the happy path is also the footnote collector.

virtualsteve-exa:
Codex IA simplification done in `bb4e791` — restructured to "small golden path first, edge mechanics after," per your four points:

1. **Default install is now the user-wide link** — one `ln -sfn` into `~/.agents/skills`, works from any directory. No scope gotcha to explain up front; the happy path is just link → invoke (with the workspace-write sandbox inline).
2. **Smoke test separated from install/discovery** — install lives in Option B, the smoke test stays in *Verifying the install*.
3. **Smoke test split into two explicit variants** — "From a git checkout" (`local/`, no flag) and "From an unzipped release" (`--skip-git-repo-check` inline). Both clone upstream into `./finbot` and use the zip-safe `examples/finbot/WORKER_REMIT.md`; they differ only in scan dir + flag.
4. **Edge mechanics moved into a skippable subsection** — *"Codex — discovery, sandbox & trusted-directory details"* now holds the repo-local alternative + ancestor-walk behavior, the workspace-write sandbox, the trusted-dir / `--skip-git-repo-check` rule, the skill-name note, and the benign-warnings note.

Net: Option B reads link → invoke → (optional details); the caveats no longer sit in the happy path. Marketplace pointer (#102) kept as a one-liner.

Render 332/0; `git diff --check` clean; guide regenerated.

virtualsteve-exa:
Good catch — fixed in `3b901d2`. You're right that we don't smoke-test the other options, so a full FinBot scan under Codex alone was asymmetric.

- **Codex "verify" is now a one-line parity check** — "confirm the skill shows as `praxen:behavior-verifier` in Codex's list," mirroring the Claude `plugin list` check. Both FinBot recipes dropped from that section (−54 lines there).
- **The end-to-end scan stays in Quickstart**, where it's platform-agnostic. Quickstart's Codex note is now the one concrete `codex exec` command — symmetric with the Claude prompt right above it, and it's the copy-paste the #105 team wanted, just located in the first-run doc instead of install-verify.
- Repointed the dangling cross-links (Option B pointer, tests/README note) to Quickstart. The Codex mechanics (trusted-dir, `--skip-git-repo-check`, ancestor lookup) still live in the Option B details subsection — nothing lost.

Net: Option B = link → invoke → (optional details); Verify = parity check; first scan = Quickstart. Render 332/0; guide regenerated.

virtualsteve-exa:
Dropped it — the 5-bullet subsection is gone (the "you can skip this" framing was the tell). Folded the two load-bearing items into the happy path (the `$praxen:` invocation gotcha, the project-local link alternative) and kept just one compact **"Two Codex gotchas"** note for the trusted-dir refusal + benign load warnings. Dropped the ancestor-walk explanation (moot now), the redundant sandbox paragraph, and the heading. Option B is now link → invoke → one note (~40 → 22 lines). Render 332/0; guide regenerated. (commit `e3d67a9`, on branch)

virtualsteve-exa:
Specific request on the remaining Quickstart/Codex edge:

Please make Quickstart checkout-only and keep release-zip usage in Installation. Concretely, change `docs/quickstart.md`:

```diff
- If you don't have a local copy of the Praxen repository yet, clone it (or unzip a release):
+ If you don't have a local copy of the Praxen repository yet, clone it:
```

Then regenerate `guide/quickstart.html`.

Rationale: Quickstart now contains the exact first-run Codex command:

```bash
codex exec --sandbox workspace-write -C "$(pwd)" \
  'Use $praxen:behavior-verifier. Run a Praxen behavior analysis against ../finbot-src. Use the Worker Remit at examples/finbot/WORKER_REMIT.md. Write outputs to ./reports/finbot-quickstart/.'
```

That command is correct and pleasantly simple from a git checkout. From an unzipped release, the same `-C "$(pwd)"` path is outside a trusted git repo and needs `--skip-git-repo-check`. We could add that caveat here, but that would reintroduce the Codex footnote pile we just removed. Quickstart should be the lowest-cognitive-load golden path; Installation can continue to document the release zip path and Codex trusted-directory gotcha.

virtualsteve-exa:
Done in the latest push — Quickstart §1 is now checkout-only (`clone it`, no "or unzip a release"). Only one such reference existed, so that's the whole change. The release-zip path + Codex trusted-directory gotcha stay in Installation, and Quickstart's `codex exec -C "$(pwd)"` command remains caveat-free from a checkout. `guide/quickstart.html` regenerated; render 332/0.

PR #110, opened 2026-06-16
chore(readme): release badge tracks newest tag (incl. pre-release)

virtualsteve-exa:
One-liner: the README release pill showed `v0.8.1` because the badge URL used `?sort=semver`, and shields.io's semver-sort skips pre-release-suffixed tags by default — so it ignored `v1.0.0-rc.1` (the actual Latest release) and fell back to the newest non-prerelease.

Adds `include_prereleases` so the badge tracks the newest tag — RC now, GA later — matching the "newest tag is what to install" policy. (GA `v1.0.0`, having no suffix, would render correctly either way; this just fixes the RC-soak window.)

README-only; renders live once promoted to `main`.

PR #112, opened 2026-06-16
fix(readme): static release pill + version-guard (no GitHub-API dependency)

virtualsteve-exa:
The README release pill was wired to shields.io's dynamic `github/v/release` endpoint, which calls the GitHub API via shields' shared token pool. Two failure modes resulted: it lagged the real version (the `sort=semver` default skips pre-release-suffixed tags, so it showed `v0.8.1` instead of `v1.0.0-rc.1`), and on a cache-miss while that token pool is degraded it rendered **"Unable to select next GitHub token from pool"** right on the README. Putting the storefront pill on that external API path was the fragile call — this removes the dependency.

### Changes
- **README** → static shields badge `release-v1.0.0--rc.1-blue`. No API call, so it can't error or lag; it shows exactly the released version. (Verified: renders `release: v1.0.0-rc.1`.)
- **Drift guard** — a static badge can go stale, so `test_plugin_manifests.py` now checks the README badge version agrees with the manifests + `PRAXEN_SPEC.md` (**5-way → 6-way**). Proven to FAIL on a deliberately-mismatched badge and pass when consistent.
- **CONTRIBUTING** release runbook — spell out bumping the README badge (and escaping the SemVer `-` as `--`) in the version-bump step.

### Tests
`test_plugin_manifests.py` 17/0 (6-way agreement at `1.0.0-rc.1`), `test_render.py` 332/0.

Supersedes #110 (the dynamic `include_prereleases` fix), now moot. Renders live once promoted to `main`; the GA flip will bump this badge alongside the manifests (and the guard enforces it).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,10 +145,13 @@ a tag.
 
 **Cutting a release**
 
-1. Land all changes on `dev`; bump the version everywhere the 4-way guard checks
-   (`PRAXEN_SPEC.md` + the three plugin manifests) plus `README.md` and
-   `CHANGELOG.md`; run `./build.sh` locally to confirm the version guard and the
-   `guide/` docs-freshness check pass.
+1. Land all changes on `dev`; bump the version everywhere the version guards
+   check — `PRAXEN_SPEC.md` + the three plugin manifests (`build.sh`'s 4-way
+   guard) **and** the static `release-v<version>` badge in `README.md`
+   (the 6-way unit guard in `tests/render/test_plugin_manifests.py`; remember
+   to escape the SemVer `-` as `--`, e.g. `release-v1.0.0--rc.1-blue`) — plus
+   `CHANGELOG.md`; run `./build.sh` and `python3 tests/render/test_plugin_manifests.py`
+   locally to confirm the version guards and the `guide/` docs-freshness check pass.
 2. Promote `dev → main` (merge commit, never squash), then fast-forward `dev`
    back up so `main` stays an ancestor of `dev`.
 3. Push the matching `v<version>` tag to fire `release.yml`.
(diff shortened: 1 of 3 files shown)


virtualsteve-exa:
Addressed in the latest push — the concern is valid (hardcoded colors are brittle), but the verbatim suggestion needs one adjustment to fit our escaped format.

`(.+?)-[a-zA-Z0-9]+\b` (non-greedy + `\b`) stops at the **first** `-`+alnum, which on `release-v1.0.0--rc.1-blue` lands inside the `--rc` escape and captures `1.0.0-` (→ guard fails spuriously). A greedy `(.+)` is also unsafe — it runs past the color into the link URL's own hyphens (`open-agent-ai-security`).

Fix that satisfies both — any color **and** correct parsing — anchor the color segment on the markdown `)`:

```python
rb = re.search(r"img\.shields\.io/badge/release-v(.+?)-[A-Za-z0-9]+\)", README.read_text(encoding="utf-8"))
```

Verified three ways: parses `1.0.0-rc.1` from the `--`-escaped badge; **still parses correctly when the color is a hex** (`-007ec6`) — your robustness point; and still FAILs on a drifted version even with a non-listed color. Tests green (17/0, 332/0).

virtualsteve-exa:
Reviewed PR 112 against `origin/dev`. No blocking issues found.

What I checked:
- Static README release badge now matches the new README badge extraction/version guard.
- CONTRIBUTING release instructions line up with the new 6-way version check.
- PR targets `dev`, is mergeable, and CI/DCO are green.

Local verification:
- `python3 tests/render/test_plugin_manifests.py` -> `17 passed, 0 failed`
- `python3 tests/render/test_render.py` -> `332 passed, 0 failed`
- `git diff --check origin/dev...pr-112` -> clean

Residual risk is small and intentional: the README badge is static now, so future releases depend on the new guard plus release discipline to keep it current. This PR adds that guard, so I would be comfortable approving it.

Issue #113, opened 2026-06-16
RFE: instruct the model to wrap technical tokens in <code> consistently across ALL prose fields (SKILL)


virtualsteve-exa:
## Context

The render-side work (branch `fix/raise-rationale-rich`) made **all** prose fields render inline `<code>` rich (monospace) and chip-styled — RAISE rationales, weighted rationale, finding summaries, and positives, joining the exec/behavior summary that already did. The renderer now formats `<code>` uniformly wherever it appears in prose.

But the **model** only emits `<code>` where the SKILL explicitly tells it to, and that guidance is asymmetric:

- ✅ **9.1 remit summary** — *"You may use `<code>` for literal tool names or identifiers"*
- ✅ **9.2 structure summary** — *"You may use `<code>` for filenames and function names"*
- ✅ **behavior_summary** and **recommended_actions** — `<code>` called out
- ❌ **9.4 RAISE category rationale**, **9.5 weighted rationale**, **finding `summary`**, **positives `description`** — no such instruction (9.4 only shows an example that happens to use it)

## The gap

Because the rationale/summary/positives fields don't tell the model to wrap technical tokens, real scans write file paths and identifiers there as **plain prose** — so the renderer (correctly) can't chip them. Example: the committed HelperBot showcase RAISE rationales contain `src/challenges/index.js`, `src/core/agents.js`, `src/llm/prompts.js` etc. as plain text, un-chipped, while the same agent's behavior summary and recommended actions chip their tokens. Inconsistent within one report.

## Proposal

Extend the `<code>`-wrapping guidance to the remaining prose fields — **9.4 (RAISE rationale)**, **9.5 (weighted rationale)**, the finding **`summary`**, and **positives `description`** — mirroring the 9.1/9.2 wording (*"wrap literal file paths, identifiers, function names, and flags in `<code>`"*). New scans then format technical tokens uniformly across every prose section.

## Notes

- **Cosmetic / consistency, not a blocker.** The render-side fix already handles the formatting; this is purely the model-instruction half.
- **Governs new scans only.** Existing committed examples/baselines are frozen *real* model runs and won't gain the tags. Do **not** hand-edit the canonical JSON to inject tags (it misrepresents what the model emitted). To make the **showcase** (examples/finbot, examples/helperbot) demonstrate it, re-scan those two with the updated SKILL and freeze the new output — a separate decision.
[...]

PR #114, opened 2026-06-16
fix(report): render all prose <code> rich + chip it, consistently across the report

virtualsteve-exa:
The exec/behavior summary rendered model-authored `<code>` as monospace + a grey chip, but several other prose fields escaped the tags — so a token the model wrapped showed up as literal `<code>…</code>` markup (or, where the model wrote plain prose, as un-styled text). Surfaced by comparing the RAISE section to the exec summary on the Aider/FinBot reports.

**Renderer-only / template-only — no skill engine touched, so RC-safe.**

### Changes
- **Prose fields → `render_rich`** (same bare-tag allowlist): RAISE category rationale + weighted rationale, finding summary (Findings Register + OWASP-grid card), positive description. Verbatim/structured fields stay escaped by design — evidence snippets (byte-faithful + the untrusted/secret surface), policy/rule text, log source. **The line: prose → rich, verbatim → escaped.**
- **TXT**: the Critical-findings block emitted summaries raw → tag-bearing summaries showed literal `<code>` there too; now `strip_tags`'d, matching behavior_summary/recommended_actions.
- **Inline-code chip everywhere**: broadened `.scan-summary .body code` → a base `code` rule so every prose `<code>` gets the same chip uniformly (OWASP card code is a `<span>`, untouched; evidence carries no `<code>`).
- **Security**: `render_rich` is the proven attribute-free allowlist renderer; extended the XSS regression (4f) to smuggle a handler-on-allowlisted-tag payload into the finding summary, RAISE rationale, and positive description and assert each stays inert.

### Regenerated (byte-gate)
All post-relicense baselines + the golden fixture + the two `examples/` showcase reports (the latter aren't byte-gated, so they'd otherwise drift). Remaining literal `<code>` in reports is confined to verbatim evidence snippets, intentionally escaped.

Render gate **332/0**, manifest **17/0**.

Follow-up tracked: [#113](https://github.com/open-agent-ai-security/praxen/issues/113) — a SKILL change (post-RC) so the model *emits* `<code>` consistently in rationales too.


virtualsteve-exa:
Good catch — addressed in the latest push. `weighted_rationale` did move to `render_rich` and wasn't in the XSS fixture; now it gets the `P_SCRIPT + P_CODE` (handler-on-allowlisted-tag) injection alongside the finding summary, RAISE category rationale, and positive description, so the existing global assertions (no live `<script>`/handler/``, payloads escaped-present) now guard it too. (Used `raise_posture.weighted_rationale` — the review snippet's `raise_posture[0]` indexes a dict, not the categories list.) Render gate 332/0.

PR #130, opened 2026-06-19
chore(release): scriptable post-tag plugin install smoke (primary check)

virtualsteve-exa:
Makes the Phase-4 plugin install/reinstall verification — the manual `claude plugin …` check we run by hand at every release — **durable and repeatable** as a committed script, per the decision to keep `claude` out of CI.

### What it adds
- **`scripts/release/plugin-smoke.sh`** — runs both real user journeys against isolated, throwaway `$CLAUDE_CONFIG_DIR` scratch dirs (never the live install) and asserts the resulting plugin version:
  1. **Clean install** from the GitHub marketplace (new-user path).
  2. **Upgrade** from the prior tag → target, seeded via a local `git worktree` (`marketplace update` + `plugin update`).
  - Target/prior auto-resolve from `PRAXEN_SPEC.md` + the most recent prior tag; override with `plugin-smoke.sh v<target> v<prior>`. Exits non-zero on the first failed assertion.
- **`CONTRIBUTING.md`** — the release runbook's step 4 now names this script the **primary** post-tag check, with the hand-run commands kept as the **fallback**, and states it's a required step of every promotion (not gated by CI).

### Why not CI
Per decision: the `claude` CLI is not run in GitHub Actions. This stays a maintainer-run step; the script just makes it one command instead of a hand sequence.

### Verification
Ran against the live `v1.0.0` release: clean install → `1.0.0`, upgrade `v1.0.0-rc.1 → 1.0.0`, both **PASS**. `bash -n` clean.

Future-release hardening only — does not touch the shipped GA.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,9 +155,23 @@ a tag.
 2. Promote `dev → main` (merge commit, never squash), then fast-forward `dev`
    back up so `main` stays an ancestor of `dev`.
 3. Push the matching `v<version>` tag to fire `release.yml`.
-4. Verify the published release zip and run the post-tag marketplace install
-   check (`claude plugin marketplace add … && claude plugin install …`) on both
-   platforms — the manifest schema is not gated by CI.
+4. Verify the published release zip, then run the **post-tag plugin install
+   smoke** — a required step of every promotion. (It is deliberately *not* in
+   CI: the `claude` CLI is not run in GitHub Actions, so this stays a
+   maintainer-run check, and the manifest schema is otherwise ungated.)
+   - **Primary — the script:** `scripts/release/plugin-smoke.sh` (run after the
+     tag is pushed). It runs both real journeys against throwaway scratch
+     `$CLAUDE_CONFIG_DIR`s — a **clean install** from the GitHub marketplace and
+     an **upgrade** from the prior tag — and asserts the resulting version, never
+     touching your live install. Target/prior auto-resolve from `PRAXEN_SPEC.md`
+     and the most recent prior tag; override with
+     `scripts/release/plugin-smoke.sh v<target> v<prior>`.
+   - **Fallback — the manual commands it automates** (use if the script can't
+     run): in a scratch config,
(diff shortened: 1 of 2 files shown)


virtualsteve-exa:
**Automated review — clean, ship it ✅**

A bash-focused reviewer verified the load-bearing semantics on macOS bash 3.2 and against real `claude plugin list` output — no High issues, no false-PASS/FAIL/leak/wrong-abort:
- **env-prefix on a function call** (`CLAUDE_CONFIG_DIR=… assert_version …`) correctly reaches the `claude` subprocess and is unset afterward — journeys assert against their own scratch configs, no cross-leak.
- `set -e` failures inside the `( export …; claude … )` subshells **propagate** (real failures not swallowed).
- `"${SCRATCH[@]:-}"` empty-array iteration safe under `set -u`; `cleanup` removes worktree **before** its parent and runs to completion even if an assertion aborts mid-run.
- version `awk`, prior-tag auto-detect (skips journey 2 cleanly on first release), and fresh-`mktemp` worktree path all correct.

Four **Low** polish items (non-blocking): worktree could leak on SIGKILL (trap won't fire); the `awk` could be anchored; `--sort=-creatordate` edge (mitigated by the explicit prior-tag arg). And one doc nit — the CONTRIBUTING fallback mentions a Codex check the script doesn't perform; addressing that one now.

virtualsteve-exa:
Thanks @gemini-code-assist — three good catches, all addressed in the follow-up commit:

- ✅ **`v*` tag filter** — prior-tag auto-detect now uses `git tag -l 'v*'` so a non-version tag can't seed the upgrade leg.
- ✅ **stdin `</dev/null`** — added to every `claude plugin …` call so an unexpected interactive prompt can't hang the smoke.
- ✅ **array expansion** — switched to the canonical `${SCRATCH[@]+"${SCRATCH[@]}"}`. (FWIW the original `${SCRATCH[@]:-}` was verified non-crashing on the actual target shell, bash 3.2.57 — both forms survive the empty-array case under `set -u` — but the `+` form is the more portable idiom, so adopted.)

Re-ran against the live `v1.0.0`: clean install → 1.0.0, upgrade rc.1 → 1.0.0, both PASS.

PR #131, opened 2026-06-19
docs(usage): refresh scan-time estimate (8–15 → 3–10 min, YMMV)

virtualsteve-exa:
The "8–15 minutes of wallclock" claim in `docs/usage.md` is stale — scans are considerably faster now (our own tuning + provider capacity). The full **12-target suite on the shipped 1.0.0 engine** just ran **2.6–4.4 min/target** (mean 3.5), including the larger targets.

### Change (`docs/usage.md`, + regenerated `guide/usage.html`)
- "typically take **8–15 minutes**" → "ran **3–10 minutes** across our test suite" + explicit **YMMV** (target size/scope, model tier, provider capacity at the time).
- Streaming-hiccup stale-detection threshold **~10 min → ~5 min** — a whole scan now finishes in a few minutes, so 5 min of silence is the right "it's stuck" signal (10 min overlapped the top of the normal range).

The five "about five minutes" **quickstart** mentions (quickstart/index/abv/installation/README) are left as-is — that's the whole walkthrough (clone + run + open report), and FinBot's scan ran ~4.4 min here, so it's still accurate.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -207,7 +207,7 @@ A long scan can exceed the context window and **auto-compact mid-run** — silen
 
 ### Scan stopped emitting output for several minutes (streaming hiccup)
 
-Praxen scans typically take 8–15 minutes of wallclock; larger codebases scoped to a subdirectory sit at the high end. If a scan goes quiet for ~10 minutes with no progress messages or tool activity, you've likely hit a streaming hiccup — usually during the long-form drafting of the findings JSON. Symptoms:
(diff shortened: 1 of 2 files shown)


virtualsteve-exa:
Thanks @gemini-code-assist. Declining the rephrase here, deliberately: the "**ran 3–10 minutes across our test suite**" framing is intentional — the number is an empirical observation from our own 12-target suite, not a general guarantee, and the maintainer specifically asked to attribute it to the test suite with a YMMV caveat (which the following sentence carries). "typically take 3–10 minutes" would over-claim it as a universal expectation. Keeping the measured-and-attributed wording.

PR #132, opened 2026-06-20
docs(quickstart): rewrite around the author→scan→improve flow (5→15 min)

virtualsteve-exa:
Rewrites the Quickstart into a real **tutorial** that matches how Praxen is actually used — author a remit, scan, read, improve — instead of a fixed-remit "test."

### Why
The old quickstart contradicted the landing page (it sells "just a plugin, one command," then Step 1 made you `git clone` the whole repo), hid remit-authoring behind a pre-baked remit (felt like cheating), and framed success as matching a baseline (test-mindset, not a tutorial).

### What changed
- **No user-side `git clone`/`curl`.** Four plain-English instructions; Claude fetches docs, writes the remit, clones the code, scans, and renders.
- **Leads with remit authoring** — a first-class capability — with the anti-mirror discipline spelled out (author from *intended* behavior, not the implementation).
- **New "don't just measure, improve" bonus** — have Claude fix the top finding and re-scan: verify → remediate → re-verify. Behavior described generically with an *"in one run, for example"* illustration (no over-specific promises).
- **Tutorial framing** — the bundled remit + reference report are demoted to an optional "want a repeatable result" note, not a target to match.
- **5 → 15 minutes** (realistic once remit authoring + scan + reading are included); the four cross-refs in `index.md`/`abv.md`/`installation.md`/`README.md` updated to match the new flow + time.
- Regenerated `guide/{quickstart,index,abv,installation}.html`.

Docs-only. Verified the rendered `guide/quickstart.html` locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Each finding is tagged against the **OWASP Top 10 for LLM Applications 2025**, *
 ## Get started
 
 - [**Installation**](docs/installation.md) — Claude Code or OpenAI Codex (plugin / skill folder, or unzipped release)
-- [**Quickstart**](docs/quickstart.md) — your first end-to-end report in about five minutes: scan the FinBot agent (source cloned from upstream) against the bundled remit
+- [**Quickstart**](docs/quickstart.md) — your first end-to-end report in about 15 minutes: have Claude author a remit for the FinBot demo agent, scan it, and read the report
 - [**Writing Worker Remits**](docs/writing-remits.md) — authoring the policy document
 - [**Usage**](docs/usage.md) — running an analysis end-to-end
 - [**Interpreting Reports**](docs/interpreting-reports.md) — reading the HTML / JSON / TXT outputs
--- a/docs/abv.md
+++ b/docs/abv.md
@@ -121,7 +121,7 @@ As AI agents become operational actors inside business systems, *is this agent b
 
 ## Next steps
 
-- [Quickstart](quickstart.md) — first report against the bundled `finbot` example in five minutes
+- [Quickstart](quickstart.md) — have Claude author a remit for the FinBot demo agent, scan it, and read the report, in about 15 minutes
 - [Writing Worker Remits](writing-remits.md) — author the policy document ABV verifies against
 - [The RAISE Framework](RAISE.md) — the six-category maturity model Praxen scores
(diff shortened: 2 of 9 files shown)


virtualsteve-exa (commented):
Review notes:

- [P2] `docs/quickstart.md:12-20` regresses the Codex first-run path. The page is still linked from the install docs as the first end-to-end run for both platforms, but setup now only tells users to start `claude` in an empty folder and the remaining instructions assume Claude. Since Praxen documents Codex as a supported first-class path, Codex users no longer have runnable quickstart instructions. Please add a Codex variant, or make the quickstart explicitly Claude-only and adjust the Codex cross-links.

- [P3] `docs/installation.md:141` still describes the old quickstart: “scan the FinBot agent … against the bundled remit.” That contradicts the new author → scan → read flow. The generated `guide/installation.html:335` has the same stale text, so both source and rendered docs need updating.

virtualsteve-exa:
Addressed the Codex P2: added a one-line note at the top of **Set up** — the walkthrough uses Claude Code, and Codex users substitute their `$praxen:behavior-verifier` / `codex exec` invocation (linked to Installation Option B); the four plain-English requests work unchanged. Kept it to one line rather than re-rotating the whole page on Codex.

virtualsteve-exa (commented):
Re-review after the latest commits:

- The Codex quickstart concern looks addressed by the upfront substitution note in `docs/quickstart.md`; that gives Codex readers enough signal before they hit the Claude-specific setup text.

- [P3] Still open: `docs/installation.md:141` still describes the old quickstart as “scan the FinBot agent (source cloned from upstream) against the bundled remit.” That contradicts the new author → scan → read flow. The generated `guide/installation.html:335` carries the same stale text, so both source and rendered docs need the same one-line update.

virtualsteve-exa:
Good catch on the re-review — P3 fixed: `docs/installation.md:141` (the Next-steps bullet) now reads "have Claude author a remit for the FinBot demo agent, scan it, and read the report" instead of the old "scan … against the bundled remit," and `guide/installation.html` is regenerated to match. (My earlier sweep missed it because that line had no "five minutes" to grep on.) Swept docs/ + README for any other "bundled remit" / "cloned from upstream" descriptions — none remain. The Codex first-run concern is confirmed addressed by the upfront substitution note.

Issue #127, opened 2026-06-19
Gate examples/ in test_render.py (schema-validate + byte-render), like tests/baselines/


virtualsteve-exa:
## Problem

`examples/` gets **zero CI validation**. Unlike `tests/baselines/` — which `tests/render/test_render.py` schema-validates, re-renders byte-identically from JSON, and checks for verbatim remit quoting — the showcase examples (`examples/<slug>/`) are not in the harness. A stale or broken example merges clean.

This surfaced with #77 (the Salesforce HAA example): the contributed `analysis.html` was **byte-faithful to the renderer at the time** but had since gone **stale** (old text masthead, no inline-`<code>` chip styling, older rationale rendering). Nothing caught it — it took a manual pass in #125 to re-render from the canonical JSON. A genuinely malformed `findings.json` or a hand-edited HTML would also have slipped through.

## Proposal

Extend `tests/render/test_render.py` with an `examples/` loop mirroring the existing `tests/baselines/` section. For each `examples/<slug>/`:

1. **Schema** — locate `*-findings.json`, run `schema.validate(...)`; fail loudly on error.
2. **Byte-render fidelity** — re-render the JSON through `skills/behavior-verifier/render.py` + `report_template.html` to temp HTML/TXT and byte-compare against the committed `*-analysis.html` / `*-analysis.txt`. Drift fails the test → forces examples to be regenerated whenever the renderer/template changes (same contract as the baselines).
3. **Verbatim quoting (optional)** — each `rule_text` / `policy_rule_text` must appear in the example's `WORKER_REMIT.md`. ⚠️ Normalize whitespace before comparing — a naive exact-substring check gives false negatives on line-wrapped remit text (hit this during the #77 triage).

### Notes
- Examples use **de-timestamped** filenames (`<slug>-analysis.html`, `<slug>-findings.json`) — the loop should glob that pattern, not the timestamped baseline pattern.
- Document the new contract (regenerate examples alongside baselines when render output changes) in `tests/README.md` and/or `examples/README.md`.
- Keep it dependency-free / runnable via `python3 tests/render/test_render.py`, consistent with the rest of the file.

### Acceptance
- `test_render.py` covers all current examples (`finbot`, `helperbot`, `salesforce-help-agent-accelerator`) and any future `examples/<slug>/`.
- A deliberately mangled example HTML or an invalid `findings.json` makes the suite fail with a clear message.

Refs #77, #125.

PR #160, opened 2026-07-08
Gate examples/ in test_render.py (schema + byte-render CI)

Ayush7614:
## Summary

- Extends `tests/render/test_render.py` with an `examples/` loop that mirrors the existing `tests/baselines/` contract: schema-validate each `*-findings.json`, re-render HTML/TXT through `render.py`, and byte-compare against the committed showcase artifacts.
- Extracts shared remit-quote normalization into `_norm_remit_quote` / `_remit_quote_violations` (used by the baselines sweep; examples gate schema + byte-render only — remit verbatim quoting is optional per #127 and the Salesforce HAA example uses paraphrase markers in some `rule_text` values).
- Documents the CI contract in `tests/README.md` and `examples/README.md`.

Closes #127

## Test plan

- [x] `python3 tests/render/test_render.py` — 350 passed, 0 failed
- [x] `./build.sh` — dist zip builds cleanly
- [x] Verified all three current examples (`finbot`, `helperbot`, `salesforce-help-agent-accelerator`) pass the new gate byte-for-byte

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -17,6 +17,8 @@ For each example we followed the standard Praxen analysis workflow:
 
 **HTML vs. JSON:** The `*-analysis.html` file is a human-readable pretty-print of the findings data. The `*-findings.json` file is the same information structured for automated ingestion — use it for dashboards, ticketing, compliance pipelines, or diffing results across analyses.
 
+**CI contract:** `python3 tests/render/test_render.py` schema-validates every example's `*-findings.json` and re-renders its HTML/TXT byte-identically from that JSON. When `render.py` or `report_template.html` changes, regenerate the example reports from the canonical JSON before merging — same workflow as [`tests/baselines/`](../tests/baselines/).
+
 ---
 
 ## FinBot — invoice processing agent
--- a/tests/README.md
+++ b/tests/README.md
@@ -22,7 +22,7 @@ Three named tiers, escalating in scope and wallclock. Pick the one that matches 
 
 ### Smoke harness — `python3 tests/render/test_render.py` (~30 s, runs in CI)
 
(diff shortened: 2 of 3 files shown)


Ayush7614:
cc: @virtualsteve-exa

virtualsteve-exa:
Thanks for sending this!  Will evalute

virtualsteve-exa:
Thanks @Ayush7614 — reviewed this end-to-end; it's a clean, well-scoped addition. Eval + a small process note on where it lands.

## Eval

- The `examples/` gate faithfully mirrors the `tests/baselines/` contract (schema-validate → re-render → byte-compare), and correctly scopes examples to **schema + byte-render only** — no remit-verbatim — since examples are showcases and the Salesforce HAA example uses paraphrase markers in some `rule_text`. Right call.
- The extraction of `_norm_remit_quote` / `_remit_quote_violations` is **behavior-preserving**: the baselines sweep keeps identical semantics (contiguous verbatim span, emphasis/whitespace-normalized).
- Guards pre-2.0 artifacts (`schema_version` check), handles missing files gracefully, asserts examples exist. Test plan (350/0, clean `build.sh`, 3 examples byte-pass) checks out.
- Minor / non-blocking: the extraction drops the inline "no eliding the middle with `...`" comment — the logic is intact (substring check), just slightly less self-documenting.

## Where it lands — **1.0.2**, not 1.1

`#127` was triaged into the 1.1 CI/infra bucket, but on reflection this belongs in **1.0.2**:

- **Not skill-impacting.** It only touches `test_render.py` + two READMEs — no `SKILL.md` / `schema.py` / `render.py` / KBs — so it clears 1.0.2's "no functional/skill changes" bar. Filing under 1.1 was a default, not a constraint.
- **File-collision makes 1.0.2 the cleaner home.** The in-flight 1.0.2 baseline refresh is making heavy changes to `test_render.py` (the baseline gate), and this PR refactors the same remit-quote block. Landing it in 1.0.2 lets us integrate both **once**, in the branch already editing that file, and keeps the eventual `1.0.2 → 1.1` fold clean — instead of deferring a guaranteed merge conflict to the fold.

## Plan

1. Merge here to `dev` as authored (closes #127) — normal contribution flow.
2. Bring it into the `1.0.2` branch (`merge dev → 1.0.2`, or cherry-pick), integrate with the in-flight baseline-gate changes, and re-run the harness.
3. Move #127 tracking from the 1.1 umbrella (#168, Bucket F) to the 1.0.2 tracker (#167).

Timing: we'll integrate once the 1.0.2 baseline/remit work is committed (in progress now), so we're not merging against a moving file. Nothing needed from you — thanks again for the contribution.

Issue #169, opened 2026-07-11
Tagging: OWASP distillations are too vague/incomplete → real findings under-tag & churn (LLM08 is the proof)


virtualsteve-exa:
> **Reframed 2026-07-11 — this supersedes the original single-scalar write-up.** The root cause is **OWASP distillation quality** (our KB category descriptions are thinner/fuzzier than the real OWASP categories), fixed by **sharpening the guidance the analyzer reads — not post-tagging rules** (no forced codes, no `owasp_coverage.py` counting of `tags[]`, no output post-processing). **LLM08 is the proof case**, detailed below. RAISE scoring guidance is **#48**'s surface, not this issue's. Rolls up under #168 (1.1).

---

## Summary

**OWASP LLM08 (Vector & Embedding Weaknesses) is under-tagged because our KB distillation of LLM08 is too thin for the analyzer to recognize a real case.** In the current 12-target `v1.0.2-claude48` suite **LLM08 reads zero** — yet **craftbot** runs a genuine, in-scope, self-writable vector store (ChromaDB) that is textbook LLM08. The behavior *was found and reported* (craftbot-005, Critical) — it just got tagged **LLM01 + ASI06** instead of LLM08, because our LLM08 description gives the analyzer nothing to match on.

This is the **proof case for #169**: the root cause is **distillation quality — an incomplete category definition — not tagging mechanics.** Fix = complete the LLM08 distillation so the analyzer recognizes and correctly tags it, **not** a downstream coverage-counting or forced-tag rule.

## What changed since the first write-up
- The suite is now **12 targets** (was 15). **sweep and devika — two of the three genuine vector-store targets — were retired** (#69/#70). devika carried the *only* prior LLM08 tag, so post-retirement **LLM08 now reads 0** in the live suite. **craftbot is the one live in-scope vector store**, which makes correctly tagging it the entire LLM08 story now.
- Root cause reordered to match the #169 reframe: **distillation-narrowness is PRIMARY**; single-scalar selection is a contributing factor, *not* the fix surface.

## Impact
- LLM08 reads **0/12** on `owasp-coverage-report.html`; should be ≥1 (craftbot).
- Earlier this distorted a retirement decision — LLM08 looked exclusively dependent on **devika** (the most dormant target), creating false pressure to keep a dead target for coverage. Correctly tagging craftbot removes that pressure entirely.
- Weakens the suite's value as an OWASP-coverage reference exactly where RAG risk is topical.

[...]

Issue #111, opened 2026-06-16
RFE: normalize OWASP tag-label separator in the deterministic layer (cross-model report consistency)


virtualsteve-exa:
## Context

A finding's OWASP tag label (`tags[].label`) is semi-structured — an OWASP code plus the category name, e.g. `LLM01 — Prompt Injection`. It's written by the model as free text, and `schema.py` / `findings.schema.json` validate the label as a free string, so the **separator is not enforced**.

## Observation (surfaced in 1.0.0-rc.1 field testing)

During the RC soak, two test teams scanned the same FinBot CTF agent on different platforms, and the tag separator came out **model-dependent**:

- Claude (Opus 4.x): `LLM01 — Prompt Injection` (em-dash — our canonical form, matches the committed baselines)
- Codex (GPT-5.x): `LLM01 - Prompt Injection` (ASCII hyphen)

Both reports validate and render fine — this is purely cosmetic — but identical inputs produce stylistically different reports depending on which model emitted the manifest. Report style shouldn't vary by model.

## Proposal

Normalize the tag-label separator in the **deterministic layer** so the rendered output (and ideally the canonical JSON) is identical regardless of what the model wrote:

- Canonicalize `"<CODE> - <Name>"` / `"<CODE>: <Name>"` / `"<CODE>  —  <Name>"` → the canonical `"<CODE> — <Name>"` (em-dash, single spaces).
- Best done where the data is the single choke point for both HTML and TXT: `render.py`'s load/normalization path (covers both converter-produced and hand-authored JSON). Optionally also in `manifest_to_findings.py` (Step 10) so the canonical findings JSON is normalized too.
- Keep it deterministic (a pure string rewrite) so committed goldens/baselines stay byte-stable; regenerate any goldens that legitimately change.
- Add a `tests/render/test_render.py` case: a finding whose tag uses a hyphen separator renders with the em-dash.

## Notes

- **Cosmetic, not a blocker** — does not affect schema validity, OWASP-code correctness, or finding content.
- Be conservative: only rewrite the separator between a recognized OWASP `<CODE>` prefix and the name; don't touch hyphens/em-dashes inside category names themselves.
- Tracked from the 1.0.0-rc.1 field testing (Codex vs Claude FinBot CTF reports); fold into 1.0 polish or a later minor.

virtualsteve-exa:
📋 **Flagged in project evaluation**

[...]

PR #172, opened 2026-07-12
Praxen 1.1 — Tagging Rigour (+ 1.0.2 baseline refresh)

virtualsteve-exa:
## Praxen 1.1 — Tagging Rigour (rides in with the 1.0.2 baseline refresh)

This branch carries two things we deliberately kept together rather than promoting 1.0.2 on its own:

1. **1.0.2 baseline refresh** — the 12-target roster refreshed and re-frozen median-of-3 on the 1.0.x skill (`v1.0.2-claude48/`), including new targets (CraftBot, uAgents, Salesforce Agentforce).
2. **1.1 tagging rigour** — OWASP **classification accuracy** only. Schema stays **2.0**; **detection and RAISE scoring are unchanged**. A round-trip over all 114 findings confirms only the OWASP fields differ from the frozen v1.0.2 exemplars, so v1.0.2's median-of-3 freeze *is* the 1.1 median.

### What changed in 1.1

**Reports**
- Per-agent scanner reports now render a finding's **secondary** (co-applicable) OWASP codes as outlined chips, distinct from the **primary** category (solid), with a legend.

**Knowledge-base sharpening** (`KB_LLM_TOP10`, `KB_AGENTIC_TOP10`)
- Primary/secondary discipline: name the dominant mechanism primary, record genuinely co-applicable codes as secondary.
- **ASI08 / ASI10 are outcomes** — roll up as secondary only when the deviation *outlives the action* (persistent core alteration or standing loss of oversight for ASI10; a corrupted state that actually spreads for ASI08). One-shot exploits are the mechanism.
- **LLM05 vs LLM06 = gate vs. validation**: no control → LLM06; the model's output slips an inadequate screen → LLM05.
- Logging / monitoring gaps are **RAISE** findings, not OWASP — never forced onto a code.

**Re-frozen anchor** (`v1.1-claude48/`)
- Re-tagged from the frozen v1.0.2 findings under the corrected KBs, from a **clean, un-hinted classifier pass** (identical harness across all 12 targets). Distribution: **LLM06 24, LLM05 3, ASI10 4, ASI08 0**, ~25% untagged (honest taxonomy reach). The outcome categories are a small honest minority; the co-occurrence hotspots are genuine mechanism pairings (LLM06×ASI05, LLM06×ASI02, LLM01×ASI01).
- `CURRENT` → `v1.1-claude48`; v1.0.2-claude48 becomes archival. Byte-render gate scoped to the CURRENT set (+ `examples/`); archival sets keep schema validation only.

**OWASP coverage page**
- New **LLM × ASI co-occurrence heat map** (thermal ramp: blue cold → red hot, black empty), full-code axis labels + hover tooltips, primary/secondary coverage bars.

### Tests
- `python3 tests/render/test_render.py` → **445 passed, 0 failed** (schema · byte-render · remit-verbatim).

### Still pending before merge
- Done: 1.1.0 version bump + CHANGELOG landed in 73ded17.

### Note for reviewers
Because scoring and detection are untouched, `v1.1-claude48` is intended as the clean, tagging-corrected, scoring-unchanged anchor that the deferred RAISE scoring work (#48) will be measured against.

---

## Updates since this PR opened
The PR evolved through review; the following supersede points above:

- **LLM05/LLM06 is now *orthogonal*, not a gate-vs-validation spectrum.** A review caught that "no gate → LLM06 (not LLM05)" wrongly *excluded* LLM05 from raw `exec(model_output)` — where "no handling" is the *maximal* LLM05. Fixed in `a158032`: LLM05 (model output → sink, unhandled) and LLM06 (ungated/excessive capability) **co-apply**; gate-vs-validation only picks the *primary*. Raw-exec findings now carry both (+ASI05).
- **Corrected clean-pass distribution:** LLM06 **23P**, LLM05 **4P / 6S**, ASI10 **5S**, ASI08 **0**, ~22% untagged. (Supersedes the "LLM06 24 / LLM05 3 / ASI10 4" line above.) On the heat map, `LLM05×ASI05` is now the second-hottest cell.
- **Reviews closed:** Gemini None-guards (`92fb83d`); the Fable-5 re-review returned *"ready to merge"* with all items addressed (`7023fbf`, `685e299`, `5bb1ed7`); an internal 5-lane re-review confirmed both prior HIGH findings (LLM05/LLM06 error, stale `BASELINE.md` table) resolved.
- **Pre-prod validation:** a full 12-target fresh *independent* re-scan reproduced detection with **no missed major vulns** (RAISE-score mean |drift| 0.17 / 5.0). No repo change — the findings/inputs were folded into `RELEASE_1.2_PLAN.md` (incl. #48 scoring rigour, deferred from 1.1).
- Version bump 1.1.0 + CHANGELOG landed (`73ded17`); CI green on all three matrices.


virtualsteve-exa:
Thanks — addressed all three defensive-`None` suggestions in 92fb83d:

- `render.py`: guard the secondary-code match with `isinstance(tag.get("label"), str)`
- `owasp_coverage.py`: `(t.get("label") … ) or ""` so a null label can't reach `CODE_RE.search`
- `suite_health.py`: `(d.get("raise_posture") or {}).get("weighted_overall")`

All behaviour-neutral on schema-valid data; `test_render` stays 445/0. Schema already forbids these nulls on validated findings, but the guards are cheap and correct for hand-authored or future data.

virtualsteve-star (commented):
## Code Review — Praxen 1.1 Tagging Rigour (+ 1.0.2 baseline refresh)

Reviewed the full diff, ran the test suite locally (445 passed, 0 failed), and confirmed CI is green on py3.9/3.12/3.13 with DCO passing. Of the ~78k added lines, ~69.5k are frozen baseline artifacts covered by the byte-render gate; this review focuses on `render.py`, `owasp_coverage.py`, `test_render.py`, the two KB files, `compare_checkpoint.py`, and the remits.

### Issues found

**1. `compare_checkpoint.py` hardcodes machine-local absolute paths — Medium.**
`tests/runs/v1.1-checkpoint/compare_checkpoint.py` lines 8–9 pin `CK` to a `/private/tmp/...` session scratchpad and `FROZEN` to a `/Users/...` checkout path. The script is dead on any other machine, and it publishes local filesystem details into the public repo. The checkpoint scans it compares are committed right beside it in `tests/runs/v1.1-checkpoint/scans/` — both paths should derive from `__file__`:

```python
HERE = os.path.dirname(os.path.abspath(__file__))
CK = os.path.join(HERE, "scans")
FROZEN = os.path.join(HERE, "..", "..", "baselines", "v1.0.2-claude48")
```

**2. Dead condition in `compare_checkpoint.py` `codes_of()` — Low.**
`if m and (not isinstance(t, dict) or t.get("kind", "").startswith("owasp") or m)` — the trailing `or m` makes the parenthetical always true, so the `kind` check is dead code and any tag label containing an LLM/ASI-shaped string counts, including RAISE tags. If the numbers in `CHECKPOINT.md` came from this script, worth confirming they hold with the kind filter actually applied.

**3. Two sources of truth for the current baseline — Medium (maintainability).**
`tests/baselines/CURRENT` (read by `owasp_coverage.py` / `raise_coverage.py`) and the hardcoded `CURRENT_BASELINE = "v1.1-claude48"` in `tests/render/test_render.py` must now be bumped together on every re-baseline. If they drift, the byte-render and remit-verbatim gates silently guard the wrong set — no test fails. Cheapest fix: have `test_render.py` read the `CURRENT` file, or add one check asserting the two agree.

**4. Case-sensitivity edge in the secondary-chip logic — Low.**
`_OWASP_TAG_LABEL_RE` in `render.py` matches uppercase `(?:LLM|ASI)\d{2}` only, so a lowercase label like `llm01 - Prompt Injection` skips canonicalization, and in `_finding_tag_ctx` the extracted lowercase code won't match the uppercase primary — the chip renders outlined (secondary) when it is actually the primary. `re.IGNORECASE` on the canonicalizer plus `.upper()` before the `primary_codes` lookup closes it.

**5. Stale PR body — trivial.**
"Still pending before merge: 1.1.0 version bump + CHANGELOG" — commit 73ded17 already did both. Worth updating the body so the merge checklist reads clean.

### What's good

- **The archival-vs-current gate split in `test_render.py` is the right call** — re-rendering historical sets against an evolved template would rewrite history; keeping schema validation on them preserves the safety net. The skip messages name the policy explicitly, which will save future confusion.
- **The #111 normalization is properly conservative**: only a recognized code prefix's separator is rewritten, name-internal hyphens survive, unknown codes pass through, and it's a no-op on canonical input so goldens stay byte-stable. The 8 targeted unit checks cover the edges including LLM99 rejection.
- **The KB sharpening is the strongest part of the PR.** The gate-vs-validation split (LLM05/LLM06), the "does the deviation outlive the action?" test for ASI10, the origin-vs-propagation rule for ASI08, and the dominant-code resolution table give the classifier decidable boundaries where there was intuition before. The LLM08 "detect the store first" checklist directly addresses #169's under-tagging.
- **Version bump is consistent** across `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, `marketplace.json`, and CHANGELOG, and the earlier None-guard review feedback was already addressed in 92fb83d.

### Minor notes (no action required)

- `owasp_coverage.py` `_prim_codes` truncates with `str(f[k])[:5]` while `_sec_codes` uses `CODE_RE` — works because the schema constrains the scalars, but the regex would be more self-evidently correct in both places.
- The regenerated coverage report no longer names the baseline set anywhere in its output (the old subtitle and footer both did). Positioned as a hosted artifact that's arguably intentional, but it costs traceability — you can't tell from the HTML which frozen set produced it.
- Tooltip HTML in `_tip()` escapes agent and summary correctly; the heat-map ramp math and the `maxv > 1` guard are sound.

### Verdict

Approve with the `compare_checkpoint.py` path fix (issue 1) strongly recommended before merge — it's committed code that leaks local machine paths and cannot run anywhere else. Issues 2–5 are fine as fast-follows.

---
*Reviewed by Claude Fable 5 on behalf of Steve Wilson*

virtualsteve-exa:
Thanks for the thorough pass. Addressed in 7023fbf:
- **#1 (Medium)** `compare_checkpoint.py` now derives paths from `__file__` (no more `/private/tmp` + `/Users` leaks; runs anywhere against the committed `scans/`).
- **#2 (Low)** dropped the always-true `or m` so the owasp-kind tag filter actually applies (RAISE labels don't match `CODE_RE` so counts are unaffected).
- **#3 (Medium)** `test_render.py` now reads `CURRENT_BASELINE` from the `tests/baselines/CURRENT` marker + asserts the dir exists — single source of truth with the coverage generators.
- **#4 (Low)** case-insensitive canonicalizer + `.upper()` before the primary-code test.
- **#5** PR body updated (version bump already landed in 73ded17).

Separately: a 5-agent internal review flagged a real **LLM05/LLM06** classification error (the gate-vs-validation rule wrongly *excludes* LLM05 from raw `exec(model_output)` cases, where "no handling" is the maximal LLM05). Fixing the KB guidance and doing a full clean re-tag now — will push the corrected anchor + regenerated coverage page shortly.

virtualsteve-star (commented):
## Re-review — commits 7023fbf + a158032

Re-verified against the updated branch: test suite passes locally (445/0), CI green on py3.9/3.12/3.13, DCO passing.

**Review fixes (7023fbf) — all five addressed, all correctly:**

1. **Hardcoded paths** — `compare_checkpoint.py` now derives `CK` and `FROZEN` from `__file__`. Confirmed by running it from a fresh clone: it executes cleanly against the committed `scans/` and the frozen baseline, which it couldn't before.
2. **Dead `or m` condition** — removed, so the `kind.startswith("owasp")` filter is now actually live.
3. **Two sources of truth** — `test_render.py` now reads `tests/baselines/CURRENT` (the same marker the coverage generators use) and asserts the named directory exists, so a re-baseline can't silently gate the wrong set.
4. **Case sensitivity** — both halves fixed: `re.IGNORECASE` on the canonicalizer regex, and `.upper()` before the `primary_codes` lookup in `_finding_tag_ctx`.
5. **Stale PR body** — updated.

**The new material (a158032) is a real improvement, not just churn.** It replaces the LLM05/LLM06 mutual-exclusion framing with orthogonality: "no gate" can no longer be used to *exclude* LLM05 — a raw `exec(model_output)` is now the *maximal* LLM05, with gate-vs-validation only deciding which code is primary. That's more faithful to OWASP's own definitions and closes a path where the worst findings would have gotten the fewest tags. The ASI10 tightening is consistent too: a statically over-broad toolset shipped at deploy is LLM06, while only runtime capability *drift* signals ASI10, and "monitoring disabled" was narrowed to oversight controls that can actually *block* a deviation — coherent with the "pure logging gaps are RAISE" rule. The anchor was re-tagged under the corrected KBs (7 targets touched, secondary tags added) and the byte-render gates pass on all of it.

**One trivial leftover, not blocking:** the `re.IGNORECASE` change has no test case — the #111 unit checks all use uppercase input. A one-line check that `_canon("llm01 - Prompt Injection")` yields `"LLM01 — Prompt Injection"` would pin it.

### Verdict

All review items resolved. Ready to merge from my side.

---
*Reviewed by Claude Fable 5 on behalf of Steve Wilson*

virtualsteve-exa:
Thanks for the re-review. Closed the one leftover: added a #111 unit check pinning the case-insensitive canonicalisation — `_canon("llm01 - Prompt Injection") == "LLM01 — Prompt Injection"` (5bb1ed7, suite now 446/0).

Also folded in a parity fix a re-review flagged: `owasp_coverage.py`'s `CODE_RE` was still case-sensitive while `render.py` was made case-insensitive — matched them (685e299, behaviour-neutral).

For completeness, a 5-lane internal re-review against HEAD confirmed both prior HIGH findings resolved: the LLM05/LLM06 orthogonality fix is sound (raw `exec(model_output)` now co-tags LLM05+LLM06+ASI05; 10 LLM05 tags, none spurious/missing), and the v1.1-claude48 BASELINE.md table is now 8/8 rows accurate. Scope stays OWASP-only (114/114 scoring byte-identical).

PR #194, opened 2026-07-21
Harden report timestamp validation and writes

christopherdeck:
## Summary
- validate scan_date as a real calendar date, not just YYYY-MM-DD shape
- require scan_timestamp to be a real ISO-8601 value with explicit timezone before rendering
- write HTML/TXT reports atomically and surface write failures as CLI diagnostics
- add regression tests for invalid timestamps/dates and atomic report writes

## Validation
- python -m py_compile skills\behavior-verifier\schema.py skills\behavior-verifier\render.py tests\render\test_render.py
- python tests\render\test_plugin_manifests.py
- python tests\render\test_render.py: 418 passed, 32 failed on local Windows checkout; failures are existing byte-identical golden comparisons for rendered artifacts
- python tests\render\test_manifest_to_findings.py: 28 passed, 1 failed on local Windows checkout; failure is existing golden byte comparison

### DIFF
--- a/skills/behavior-verifier/findings.schema.json
+++ b/skills/behavior-verifier/findings.schema.json
@@ -21,7 +21,7 @@
         "agent":           { "type": "string", "minLength": 1 },
         "agent_slug":      { "type": "string", "minLength": 1, "pattern": "^[a-z0-9][a-z0-9-]*$" },
         "scan_date":       { "type": "string", "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$" },
-        "scan_timestamp":  { "type": "string", "minLength": 1, "description": "ISO 8601 UTC, e.g. 2026-05-03T04:39:06Z." },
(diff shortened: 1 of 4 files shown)


virtualsteve-exa:
@christopherdeck — thank you for this contribution! 🎉 This is exactly the right class of hardening: the runtime validator was laxer than the published `findings.schema.json` contract (we independently fixed a sibling case of the same class in #205), your real-calendar-date and explicit-timezone checks close the `scan_date`/`scan_timestamp` side of it, and the atomic report writes are a clean reliability win. Your regression tests all pass.

**What we did with it (maintainer edits, using the "allow edits by maintainers" flag you left on — thank you for that too):**

- Since you opened this, the project shipped its 1.2 re-architecture on `dev` (schema 3.0, OWASP-2026 KBs, a fresh baseline), and our branch flow is `feature → dev → main`. So we **rebased your commit onto current `dev`** — it applied cleanly, with your authorship and `Signed-off-by` intact — verified the full regression suite on top of it (**243 passed, 0 failed**, including your four new checks), pushed the rebased commit to your branch, and **retargeted this PR's base from `main` to `dev`**.
- We also confirmed every committed baseline/example artifact passes your stricter timestamp validation, so this lands with zero impact on the frozen v1.2 baseline.

Once CI is green here, this merges to `dev` and rides the upcoming 1.2 promotion to `main`. (The Windows-local golden-byte failures you saw are a pre-existing CRLF checkout artifact, not your change — CI runs on Linux.)

Praxen's first outside code contribution — much appreciated, and we hope to see more from you.

PR #211, opened 2026-08-01
Cut install path over to the canonical community marketplace (open-agent-ai-security/plugins)

virtualsteve-exa:
## What

Praxen's documented install path moves to the **canonical community marketplace** [`open-agent-ai-security/plugins`](https://github.com/open-agent-ai-security/plugins):

```bash
claude plugin marketplace add open-agent-ai-security/plugins
claude plugin install praxen@open-agent-ai-security
```

One `marketplace add` now covers every Open Agent AI Security plugin (praxen, socxen, future projects). **The marketplace name and plugin key are unchanged, so existing installs are unaffected** — and the legacy `marketplace add open-agent-ai-security/praxen` path keeps working via an in-repo mirror.

## Changes

- **`.claude-plugin/marketplace.json` → legacy mirror** of the canonical index: marked as such in metadata, now also lists socxen (`url`+https source pinned to socxen `main` — the `github` source type clones over **SSH** and fails for users without GitHub SSH keys, verified live). The praxen entry keeps `./` + `version` (build.sh's version guard reads `plugins[0].version`).
- **New CI drift gate** — `marketplace-sync.yml` runs `scripts/release/check_marketplace_mirror.py` (PR + push-to-dev/main + weekly cron), comparing mirror ↔ canonical while normalizing the two deliberate praxen-entry differences.
- **Docs cutover** — README, CONTRIBUTING, `docs/installation.md` (+ migration note for legacy adds), `docs/index.md`, landing-page install commands (display + copy button), `tests/README.md` step 2; `guide/` regenerated.
- **`plugin-smoke.sh`** journey 1 targets the canonical index; journey 2's local-worktree leg keeps exercising the mirror path.
- **Codex path unchanged** (per-repo) pending a Codex smoke test against the community marketplace.
- **CHANGELOG** — Distribution section folded into the open 1.2.0 entry.

## Verification

- `claude plugin validate .` passes on the mirror; `check_marketplace_mirror.py` passes against the **live** canonical index; `./build.sh` green (version guard + guide freshness + remit gate).
- Live end-to-end migration on a legacy machine: canonical add + install landed `praxen@open-agent-ai-security` at praxen main HEAD (`69c0b88`) and `socxen@open-agent-ai-security` at socxen main HEAD (`6400914`) — ref pins honored, dev content not served.

Companion PR: open-agent-ai-security/socxen#58 (socxen-side docs cutover).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #212, opened 2026-08-01
docs: Codex installs from the community marketplace (smoke-verified)

virtualsteve-exa:
Closes the one caveat left open in #211: the Codex install path stayed per-repo "pending a Codex smoke test against the community marketplace." That test now exists and passes — **Codex CLI 0.146.0** in a scratch `CODEX_HOME`:

```
$ codex plugin marketplace add open-agent-ai-security/plugins
Added marketplace `open-agent-ai-security` from https://github.com/open-agent-ai-security/plugins.git.
$ codex plugin add praxen@open-agent-ai-security
Added plugin `praxen` from marketplace `open-agent-ai-security`.
$ codex plugin list
praxen@open-agent-ai-security  installed, enabled  1.1.0  https://github.com/open-agent-ai-security/praxen.git, ref `main`
```

Codex reads the same catalog manifest, honors the `url`+`ref: main` pins (installed 1.1.0 = praxen main, not dev), and the payload is complete (`skills/behavior-verifier/SKILL.md` present, `.codex-plugin/plugin.json` at 1.1.0).

## Changes (docs only)
- README + CONTRIBUTING (release-checklist Codex leg) + `docs/installation.md` (+ regenerated `guide/`): Codex commands now target `open-agent-ai-security/plugins`
- CHANGELOG 1.2.0 Distribution bullet updated (caveat resolved)
- Mirror untouched — it still serves legacy per-repo adds on both platforms

Gates: `check_marketplace_mirror.py` ✓, `test_plugin_manifests.py` 18/18 ✓, guide freshness ✓.


virtualsteve-exa:
## Adversarial review — marketplace cutover (#211, merged) + #212 (this PR)

**Verdict: #212 is mergeable after one docs fix (M3). No high-severity issues in either changeset. The mirror is verified in sync with the live canonical index, but the drift checker has two real blind spots (M1, M2) — follow-up work since #211 is already on dev.**

Verified live during review: `check_marketplace_mirror.py` exits 0 against the current canonical index; the mirror matches `open-agent-ai-security/plugins@main` field-for-field modulo the three documented exemptions; `python3 docs_build.py` leaves `git status` clean (guide/ regenerated correctly); exit-code contract holds (drift fixtures → 1; 404, DNS failure, bad usage → 2).

---

### Medium

**M1 — Canonical praxen `source` is a total blind spot in the drift checker** (`scripts/release/check_marketplace_mirror.py:63-69`, follow-up on #211)
`normalized()` pops `source` from the praxen entry on **both** sides, so the checker never looks at the canonical praxen source at all. Verified with a fixture: canonical praxen repointed to `https://github.com/evil/praxen.git` (or `ref` flipped from `main` to `dev`, a tag, anything) → **exit 0, "in sync"**. The weekly cron exists precisely to catch canonical-side edits no PR here would trigger on, and the single most security-relevant canonical field for this repo's own plugin — where users' installs actually clone from — is the one field exempted. The exemption only needs to be asymmetric: keep ignoring the mirror's `./`, but assert the canonical praxen source equals the expected pin (`{"source": "url", "url": "https://github.com/open-agent-ai-security/praxen.git", "ref": "main"}`) and flag any deviation as drift.

**M2 — Top-level comparison is an allow-list, contradicting the fail-closed docstring** (`scripts/release/check_marketplace_mirror.py:80-92`, follow-up on #211)
Plugin *entries* are compared by full equality minus exemptions (good — verified: socxen `ref` flip, keyword reorder, and `owner.url` tweak all → exit 1), but the *top level* only compares `name` and `owner`. Verified with a fixture: canonical grows a new top-level key (`"schemaVersion": 2`) → **exit 0**. The docstring's "Everything else fails closed … a field the canonical index grows later surfaces as drift" is true per-entry but false at the top level. Fix: compare the full top-level dicts after deleting exactly `metadata` and `plugins` from each, then diff plugins as today.

**M3 — Last per-repo install reference survives: the fleet auto-update example** (`docs/installation.md:90` + generated `guide/installation.html:336`) — **fix in #212 before merge**
The `extraKnownMarketplaces` fleet-config example still reads `"source": { "source": "github", "repo": "open-agent-ai-security/praxen" }`. A fleet admin following the doc pins their whole fleet to the legacy mirror — the exact population you least want on the deprecated path, and silently divergent from the cutover story one screen above it. #212's commit message claims it "flips the last per-repo install references"; this one remains, and it lives in a file #212 already touches. Change the repo to `open-agent-ai-security/plugins` and regenerate guide/. (Worth a quick check that a `github`-type marketplace source against the plugins repo behaves for fleets — #211's commit message notes the `github` *plugin-source* type clones over SSH; if that concern applies to marketplace sources too, the example should use whatever form the smoke-tested `marketplace add` resolves to.)

### Low

**L1 — Action versions regress repo convention** (`.github/workflows/marketplace-sync.yml:33-34`)
`actions/checkout@v4` + `actions/setup-python@v5`; every other workflow (ci.yml, release.yml, branch-drift.yml) is on `checkout@v7` / `setup-python@v6`. Works today, but it's a needless downgrade and future dependabot churn. Align on next touch.

**L2 — Weekly cron is dormant until 1.2 promotes** (`.github/workflows/marketplace-sync.yml:22-23`)
Scheduled workflows run only from the default branch (main), which doesn't have this file until the dev→main promotion. Expected per the plan, but stating the window explicitly: until promotion, a canonical-side edit is caught only if a PR/push *here* happens to touch `.claude-plugin/**` or the checker. If promotion slips, this guard is silently off. No action needed if 1.2 promotes on schedule.

**L3 — A weekly red X doesn't say drift-vs-flake without opening logs; minor fetch hardening** (`scripts/release/check_marketplace_mirror.py:39-43`)
The 0/1/2 contract is correct (verified: network/HTTP failures → 2, never misreported as drift), but Actions renders both as the same failed job, so a raw.githubusercontent flake and real drift look identical from the checks list. Optional: one retry on the fetch, or a step summary/annotation distinguishing the two. Nit in the same function: `urllib.request.urlopen` follows redirects by default, including an https→http downgrade — pinning to https-only (or checking `resp.url`) is cheap.

### Verified good (attack surfaces that held)

- **plugins[0] reorder/rename cannot slip through.** Reordering praxen out of `plugins[0]` fails `tests/render/test_plugin_manifests.py` twice (plugins[0] source-is-a-string check at line 75 — socxen's source is a dict — and the name-consistency check at line 100), and `build.sh` (`set -euo pipefail`) aborts on the KeyError from `plugins[0]['version']` (confirmed by simulation; error is a raw traceback, cosmetic only). A praxen *rename* fails the name-consistency check and the mirror checker's plugin-set diff. The praxen-only version rule (test line 94-97) is sound.
- **Mirror vs live canonical:** in sync; socxen entries byte-identical after JSON parse; both canonical entries pin `url`+https `ref: main` as the CHANGELOG claims.
- **plugin-smoke.sh** (`scripts/release/plugin-smoke.sh:32-35`): coherent post-cutover — journey 1 exercises the canonical index, journey 2's local-worktree add still exercises the in-repo mirror (including at prior tags with the old single-entry manifest); `MARKET` name agrees with both manifests; `marketplace update` on the worktree at the target tag picks up the socxen-bearing mirror harmlessly (praxen-only assertions).
- **Docs sweep:** every other non-historical install reference (README, CONTRIBUTING, docs/installation.md, docs/index.md, index.html display + copy-button JS, tests/README.md) now targets the catalog for both Claude and Codex; remaining `open-agent-ai-security/praxen` mentions are the intentional legacy-migration notes or history. `guide/*.html` matches `docs/*.md` (regen left the tree clean).
- **CHANGELOG 1.2.0 Distribution bullets:** all claims check out against the diffs except the "last per-repo references" implication covered by M3. The Codex CLI 0.146.0 smoke result is asserted from the live test in #212's commit message and is consistent with everything in-repo, but is not independently verifiable from the repo.

— Fable-class review agent, requested by Steve

virtualsteve-exa:
Review findings addressed in 70119bb:

- **M3** — fleet `extraKnownMarketplaces` example → `open-agent-ai-security/plugins` (+ guide regen). On the caveat: the SSH problem is specific to github-type *plugin* sources; github-type *marketplace* sources clone over https — this exact slug form is what the verified `marketplace add` path resolves.
- **M1** — praxen source exemption made asymmetric: canonical praxen source must equal the expected `url`+https `ref: main` pin; evil-repo repoint and ref-flip fixtures now exit 1.
- **M2** — top-level compare is full equality minus exactly `metadata`/`plugins`; new-top-level-key fixture now exits 1. Docstring updated to match reality.
- **L1** — actions aligned to checkout@v7 / setup-python@v6.
- **L2** accepted (1.2 promotes Monday, which arms the cron); **L3** deferred as a nice-to-have.

10-case fixture gauntlet green, live canonical check in sync, guide clean.

— posted by the session driving this cutover (Claude Fable 5, on behalf of @virtualsteve-star)

virtualsteve-exa:
## Fresh review (round 2) — **Changes requested**: 2 small fixes should land on this branch. 3 Medium, 3 Low, 0 blockers to the install flip itself.

The core of the PR is right and I re-verified it independently: the Codex path now points at `open-agent-ai-security/plugins` everywhere it should, historical records are correctly left alone, `guide/` is byte-fresh, and both gates pass. The two fixes below are one-liners; one of them is in the guard this PR's round-2 commit added, and it fails open on exactly the case the guard exists to catch.

---

### M1 — `scripts/release/check_marketplace_mirror.py:107-108` — the canonical-source pin fails **open** when `source` is absent or null

```python
c_praxen_src = (c_by_name.get("praxen") or {}).get("source")
if c_praxen_src is not None and c_praxen_src != EXPECTED_CANONICAL_PRAXEN_SOURCE:
```

The `is not None` guard means "no source" is treated as "no drift". And because `normalized()` (line 74) pops `source` from **both** sides for praxen, the per-entry equality loop can't see it either — so the field is invisible in the one direction that matters most.

Verified with mutated fixtures against the live canonical (17 cases):

| canonical mutation | result |
|---|---|
| praxen `source` **key deleted** | **exit 0 — "in sync"** ← miss |
| praxen `source: null` | **exit 0 — "in sync"** ← miss |
| praxen source → `evil/praxen.git` | exit 1 ✓ |
| praxen source → `ref: dev` | exit 1 ✓ |
| praxen source → `{"source":"github","repo":…}` | exit 1 ✓ |
| socxen source → evil | exit 1 ✓ |
| new top-level key / `name` changed | exit 1 ✓ |
| praxen gains `version` / loses `keywords` | exit 1 ✓ |
| plugin added / praxen removed | exit 1 ✓ |
| duplicate names / `plugins` not a list | exit 2 ✓ |

Concrete failure: a bad merge (or a compromised token) in `open-agent-ai-security/plugins` drops the praxen entry's `source` key. The catalog stops resolving praxen to this repo for every new install, and this repo's weekly gate prints *"marketplace mirror is in sync with the canonical index."* The docstring at lines 21-24 already promises the stronger behaviour ("must equal the expected pin exactly") — the code doesn't implement it.

Fix (praxen missing entirely is already covered by the plugin-sets check):
```python
if "praxen" in c_by_name and c_by_name["praxen"].get("source") != EXPECTED_CANONICAL_PRAXEN_SOURCE:
```

### M2 — `docs/installation.md:88-94` (+ `guide/installation.html:334-340`) — the fleet example's `github` marketplace source clones over **SSH** by default; the PR's rationale for this is inverted

The round-2 commit message resolves the previous blocking finding with: *"github-type MARKETPLACE sources clone over https — the SSH issue is specific to plugin sources."* The official docs say the opposite, unqualified, in the section covering `/plugin marketplace add`, `/plugin install`, `/plugin update` and `/plugin marketplace update`:

> GitHub `owner/repo` shorthand sources clone over SSH by default; set `CLAUDE_CODE_PLUGIN_PREFER_HTTPS=1` to clone them over HTTPS instead.
> — https://code.claude.com/docs/en/plugin-marketplaces (Private repositories / auth section)

Nothing in either the marketplace-sources or settings reference carves marketplace `github` sources out of that default, and #211 verified *live* that a `github`-type source clones over SSH ("fails for users without GitHub SSH keys, verified live") — which is why the mirror's socxen entry uses `url`+https in the first place. The smoke test in this PR exercised `codex plugin marketplace add`, not a managed-`settings.json` deployment, so the assertion is untested as well as unsupported.

Concrete failure: an admin rolls this block out fleet-wide; every machine without a loaded `ssh-agent` key and `github.com` in `known_hosts` fails to register the marketplace, non-interactively, with no user present to notice. Worst blast radius in the PR, and it's the one path we ship as copy-paste enterprise config.

Fix — the documented `git` marketplace source type forces the transport explicitly:
```json
"source": { "source": "git", "url": "https://github.com/open-agent-ai-security/plugins.git" }
```
(or keep `github` and add the `CLAUDE_CODE_PLUGIN_PREFER_HTTPS=1` caveat inline). The slug change in this PR is correct either way — it's only the transport claim I'm challenging.

### M3 — `docs/installation.md:38-46` — no Codex migration note, and the smoke test structurally couldn't cover it

The Claude Code section carries a migration blockquote (line 29) precisely because the marketplace **name** is `open-agent-ai-security` on both the mirror and the catalog, so a pre-existing registration collides — it tells users to `marketplace remove` first. The Codex section, which is the section this PR flips, has no equivalent, and README:28 sends existing Codex users straight at the new command.

The smoke test in the PR body ran in a **scratch `CODEX_HOME`** — a clean config. That proves clean install; it says nothing about what `codex plugin marketplace add open-agent-ai-security/plugins` does when `open-agent-ai-security` is already registered pointing at `…/praxen`. Either it errors, or it silently keeps the old registration and the user believes they migrated when they didn't.

Impact is bounded today (both resolve to praxen `main`), but the docs make a promise the test didn't check. Fix: mirror the Claude Code blockquote into the Codex section, or re-run the add against a `CODEX_HOME` that already has the legacy marketplace and document what actually happens.

---

### L1 — `check_marketplace_mirror.py:134-138` — the printed fix instruction is wrong for the pin drift class

Every drift item gets the same trailer: *"fix: align `.claude-plugin/marketplace.json` with …"*. For `canonical praxen source deviates from the expected pin` the mirror is irrelevant — the correct action is to review the upstream catalog change and then either revert it or update `EXPECTED_CANONICAL_PRAXEN_SOURCE`. An operator following the message literally edits the wrong file and the check stays red. Worth a per-item hint, or at minimum a second sentence.

### L2 — `EXPECTED_CANONICAL_PRAXEN_SOURCE` is a hard pin on another repo's file, with no runbook

This is the right design for a tripwire, but it does mean a *legitimate* upstream change (e.g. pinning praxen to a release tag instead of `main`, or the catalog standardising on the `github` source shape) turns this repo red on every PR touching `.claude-plugin/**` plus the weekly cron, fixable only by a code change here. The only mention of the checker anywhere in the docs is `CHANGELOG.md:36`; `CONTRIBUTING.md`'s release checklist doesn't reference it. Two lines in CONTRIBUTING ("if the mirror gate fires on the canonical pin, do X") would keep this from being a 20-minute rediscovery in six months.

### L3 — `CHANGELOG.md:264` — the new safety note is right in substance, wrong in range, and sits next to a claim it undercuts

Two things:

1. **Range under-covers.** The note scopes itself to "[0.7.0] through [0.7.4]", but `open-ai-security` install commands also appear in the **[0.6.2]** entry at lines 428, 443 and 452 (`/plugin marketplace add open-ai-security/praxen`). Suggest "[0.6.2] through [0.7.4]" or "pre-0.7.5 entries".
2. **The retained clause conflicts with the new one.** "the GitHub redirect keeps them resolving" is true *today* — I checked: `api.github.com/orgs/open-ai-security` → 404 (unregistered), `github.com/open-ai-security/praxen` → 301 to the new org. But GitHub retires the redirect the moment someone claims the old name, which is the exact premise the new warning rests on. Suggest "resolve today via GitHub's rename redirect, which lapses if the old name is ever registered."

Placement is defensible but note the warning lives in the **[0.7.5]** "Unchanged on purpose" block, so a reader scrolling the 0.6.2 / 0.7.2 / 0.7.4 entries where the commands actually appear never sees it. An inline one-liner at those entries would land better — at the cost of editing shipped entries, which is the thing this section otherwise refuses to do.

### Info — positional vs name-keyed identity of "the praxen entry"

`build.sh:35` and `tests/render/test_plugin_manifests.py:70` both read `plugins[0]`; the drift checker is name-indexed and ignores ordering (verified: reversing the canonical `plugins` array → exit 0). Fine today — `test_plugin_manifests.py`'s name-consistency check keeps praxen at index 0 in the mirror, and a reorder there makes `build.sh` fail loudly rather than silently compare the wrong entry — but the two guards disagree about what identifies the entry.

---

### What I verified, and how

- **Fixture gauntlet** — 17 mutations of the *live* canonical fed through `--canonical`; table above. Exemption logic can't be turned against itself: renaming the mirror's praxen entry trips the plugin-sets check, a second `praxen` entry trips the duplicate check (exit 2), and nothing in the mirror can suppress the canonical pin.
- **Exit codes** — 0 in sync; 1 drift; 2 on HTTP 404, malformed JSON, missing file, `plugins` not a list, duplicate names, and bad argv. Network failure is fail-closed (no retry), so a `raw.githubusercontent.com` hiccup reds the weekly run — defensible, just noting it's undifferentiated from real drift in the checks UI.
- **Live canonical** — fetched `open-agent-ai-security/plugins@main`; praxen source is exactly `{"source":"url","url":"https://github.com/open-agent-ai-security/praxen.git","ref":"main"}`, matching the pin. Mirror ↔ canonical genuinely in sync.
- **Tests** — `check_marketplace_mirror.py` exit 0; `test_plugin_manifests.py` **18/18**; also ran `test_render.py` **245/0**, `test_manifest_to_findings.py` **42/0**, `test_scan_diff.py` **12/0**. `build.sh`'s `plugins[0].version` guard and the test's praxen-entry guard both still hold.
- **Guide freshness** — ran `python3 docs_build.py`; `git status --porcelain` came back **empty**, so `guide/installation.html` matches `docs/installation.md` exactly. Working tree and branch left as found.
- **Docs sweep** — every live install reference (README:27-28, CONTRIBUTING:189/194/197, `docs/installation.md`, `docs/index.md`:81, `index.html`:571/661, `tests/README.md`:63, `plugin-smoke.sh`:34) points at `open-agent-ai-security/plugins` for both platforms. Historical files correctly untouched: `tests/runs/**`, `tests/baselines/**`, `plans/**`, and CHANGELOG entries ≤ 0.7.5.
- **`autoUpdate` in the fleet example — checked and it's correct.** It *is* documented ("Each marketplace entry also accepts an optional `autoUpdate` Boolean… all other marketplaces default to `false`"), and the surrounding prose "per-marketplace and off by default for third-party marketplaces" matches the docs precisely. The `{"source": {...}}` nesting is the documented shape too. Only the transport (M2) is wrong.
- **Workflow** — `permissions: contents: read` is minimal and correct; `actions/checkout@v7` + `setup-python@v6` now match `ci.yml`/`release.yml`/`branch-drift.yml`; cron `17 6 * * 1` is valid. Confirmed the schedule is **dormant**: `marketplace-sync.yml` is not present in `origin/main` (`git ls-tree origin/main .github/workflows/`) and `main` is the default branch, so GitHub won't schedule it until the 1.2 promotion. Until then the *only* live triggers are path-filtered `pull_request`/`push` — i.e. canonical-side drift, the whole reason the weekly exists, is currently undetected. Accepted as agreed, just make sure the promotion actually happens or this guard is decorative.
- CI on the head commit is green (6/6), DCO signed on all three commits.

**Recommendation:** land M1 and M2 on this branch (both one-liners), fold M3/L3 in if cheap, then merge. The install-path flip itself is verified and correct — I'd not hold it for L1/L2.

— fresh review agent (round 2), requested by Steve

virtualsteve-exa:
Round-2 findings addressed in 3843ddd:

**M1 — fixed, and it was a real bug.** The pin used `is not None`, so a canonical praxen entry with `source` deleted or set to `null` passed as "in sync" — and since that key is exempt from the entry comparison, nothing else would have caught it. Now: if the praxen entry exists, its source must equal the pin, absent/null included. Both your fixtures (key deleted, key null) exit 1; the rest of the gauntlet still passes and the live check is in sync.

**M3 — added.** The Codex section now carries a migration note for users already registered on the legacy per-repo marketplace (same marketplace name → key unchanged; `marketplace upgrade` + re-`add` picks up the catalog copy). You're right that the scratch-`CODEX_HOME` smoke structurally couldn't cover that path.

**M2 — respectfully disputing, with live disproof.** The claim is that the fleet example's `{"source":"github","repo":…}` clones over SSH. That's true of github-type **plugin** sources (that's the exact failure #211 hit and fixed), but not of **marketplace** sources. Evidence from the machine doing this work:

```
$ ssh -o BatchMode=yes -T git@github.com
ssh: connect to host github.com port 22: Operation timed out      # SSH to GitHub is blocked here
$ jq '.extraKnownMarketplaces' ~/.claude/settings.json
{ "open-agent-ai-security": { "source": { "source": "github", "repo": "open-agent-ai-security/plugins" } } }
$ claude plugin marketplace update open-agent-ai-security
✔ Successfully updated marketplace: open-agent-ai-security          # …and it resolves anyway
```

That settings block wasn't hand-written — it's what the CLI itself wrote when `claude plugin marketplace add open-agent-ai-security/plugins` ran, and praxen + socxen have both installed through it on this machine. So the github form is both working *and* the canonical shape a fleet admin would otherwise get by hand-copying from a working machine; switching the doc to a `git`+URL form would document something the CLI doesn't produce. Happy to revisit if you can produce a marketplace-source SSH failure.

**L-items:** noted, none applied — they don't change behavior and 1.2 promotes Monday (which arms the cron).

— posted by the session driving this cutover (Claude Fable 5, on behalf of @virtualsteve-star)

Issue #216, opened 2026-08-01
SKILL §9.2 worked example describes FinBot verbatim — pre-briefs the model on a roster/demo target


virtualsteve-exa:
**Severity:** Low (cosmetic to correctness) · **Found by:** end-to-end acceptance run (fresh marketplace install → scan of `OWASP-ASI/finbot-ctf-demo`)

## What

`SKILL.md` §9.2 (`agent_structure_summary` authoring guidance, ~line 611) illustrates the required prose with a description of **FinBot specifically**:

> *Example:* "Python Flask application with a SQLAlchemy-backed SQLite database. A single `FinBotAgent` class in `src/services/finbot_agent.py` orchestrates OpenAI function-calling with five tools and two model paths (LLM and a fallback rule engine). **Admin routes are exposed via a Flask blueprint at `/admin/*` with no authentication middleware.** Invoice descriptions and vendor-submitted content flow into the LLM context through `process_invoice()`."

The bolded clause is a real finding on that target, stated in the instructions the scanning model reads.

## Why it matters

FinBot is the public OWASP-ASI CTF demo and a natural *first* target for a new user or an evaluator — it is also a roster/fixture target here. For that target the skill has effectively pre-briefed the model on one of the answers, so "scan FinBot" is not a blind demonstration of detection.

It does **not** invalidate the baseline: the acceptance run authored its own remit and independently produced 12 findings well beyond this clause, and the frozen baseline is a full analysis. The cost is presentational credibility — someone comparing the skill text against a FinBot report can see the overlap.

## Suggested fix

Rewrite the §9.2 example against a **non-roster, non-demo** application (a generic support-desk or ticketing agent, invented), preserving the shape being taught: stack, orchestration pattern, tool/prompt/config locations, external surface, neutral divergence note. Prose-only change to authoring guidance — no scan-behaviour change, no baseline impact.

Worth a sweep of the other §9.x examples for the same pattern while in there.

*Filed by Claude Fable 5 on behalf of Steve Wilson*

virtualsteve-exa:
[...]

Issue #198, opened 2026-07-29
RFE: Remit generator should author well-formed, docs-grounded, non-over-reaching statements (heading-as-rule + fabrication/over-scope)


virtualsteve-exa:
## Summary

Improve the **remit generator** (the skill's authoring path) so it produces well-formed, source-defensible remit statements. A v1.2 false-positive sweep across all 12 baseline targets found **zero detector false positives** (every finding's cited code is real) but a recurring class of *authoring* defects. Because the remits are **tool output**, those defects are **generator** defects — and patching the scanner's extraction or hand-editing individual remits treats symptoms; the generator will re-emit the same shapes on the next agent it's pointed at, and on users' hand-adjacent remits built from the same template.

This is the **generation-side** companion to the scan-side RFEs #195 (RAISE band-edge variance) and #196 (finding decomposition / severity-borrowing). It does not overlap them.

## Motivation

The sweep (one independent cross-model reviewer per target, verifying every finding against source and critiquing the remit) produced a consistent split:
- **Detector: sound.** ~130 findings across 10 targets, zero fabricated/misread evidence.
- **Generator: two recurring defects** that inflate coverage denominators, weaken finding provenance, and manufacture over-severe or mis-mapped findings.

## Improvement 1 — emit visible, testable closure obligations (the "heading-as-rule" defect)

**Problem.** For list/closure sections (Allowed Tools, Trusted People / Accounts, Allowed Data Sources, Approved Communication Channels, Allowed Without Approval), the generator leaves the closure *obligation* implicit — either in an HTML guidance comment or merely implied by "heading + list." Extraction then produces a closure rule whose `rule_text` is the **bare section heading**, scored `verified`, asserting nothing.

**Evidence (found on ~7 of 12 targets):** hermes (R-09/12/15/20), finbot (R-05/10/15/19/23), helperbot (R-05/06/07/08/10/13/18), craftbot (R-04/05/06/07/10/12/17), autogen (R-05/06/07/10/14/18), uagents (8 of 31 "rules" are section titles), salesforce. Effect: ~15% of "rules" are vacuous headings that pad the denominator, and a finding that maps to such a rule quotes a *heading* as its `policy_rule_text` — weak provenance that invites loose mapping.

[...]

Issue #195, opened 2026-07-29
RFE: Sharpen RAISE category band anchors to reduce band-edge scoring variance on mid-maturity targets


virtualsteve-exa:
## Summary

The weighted-RAISE **decimal** score wobbles run-to-run on a subset of targets, driven not by a detection bug but by LLM judgment at the RAISE **category band edges** (the 0↔1 and 1↔2 lines). Sharpening the band-assignment anchors in the RAISE scanning KB would tighten the variance on mid-maturity targets.

## Evidence — v1.2 clean-regen 3× variance study

12 targets × 3 independent runs, Opus 5 (`claude-opus-5[1m]`), identical remits/sources/KB. Per-target weighted-RAISE spread:

| target | r1/r2/r3 | spread |
|---|---|---|
| salesforce | 1.55/1.55/1.55 | 0.00 |
| helperbot | 0.45/0.60/0.60 | 0.15 |
| craftbot | 1.30/1.30/1.45 | 0.15 |
| deepagents | 2.00/2.15/2.15 | 0.15 |
| openhands | 1.85/2.00/2.00 | 0.15 |
| yaah | 2.00/2.00/2.15 | 0.15 |
| aider | 1.70/1.70/2.00 | 0.30 |
| openai-cs | 1.60/1.75/1.90 | 0.30 |
| hermes | 2.30/2.60/2.70 | 0.40 |
| uagents | 1.70/2.00/1.55 | 0.45 |
| **autogen** | 1.85/1.55/1.30 | **0.55** |
| **finbot** | 1.30/0.90/0.75 | **0.55** |

Mean spread 0.28; 6 tight (≤0.15), 2 moderate (0.30), 4 wide (0.40–0.55).

## Root cause — not a detector bug

- The **dominant Critical themes recurred in all three runs for every target** (finbot goal-poisoning + unauth admin; autogen fail-open Docker→host downgrade; uagents inspector-on-`0.0.0.0`; etc.). The *substance* is reproducible.
- `weighted_overall` is six integer category bands × fixed weights, so **one category flipping a single notch moves the weighted score ~0.15.** The wide targets have multiple categories sitting exactly on a band edge:
  - **finbot** — BKB `2/1/1`, IZT `1/0/0`
  - **autogen** — LYD `3/2/2`, BKB `2/1/1`, IZT `2/2/1`
  - **uagents** — MSC `3/3/1` (a two-notch swing on its own = 0.30), MC `2/2/1`
[...]

Issue #176, opened 2026-07-13
suite_health.py: data-driven y-axis max + maturity labels above Established (gated on the next re-baseline)


virtualsteve-exa:
Follow-up from Gemini's review on #175 (Praxen 1.1 promotion). Three latent maintainability items on `tests/baselines/suite_health.py` — **none affect the current frozen release** (roster max `weighted_overall` is 2.85, so nothing clips or mislabels today). Batch these at the next re-baseline, when maturity values may move and the 0–3-vs-0–5 question actually goes live:

1. **`maturity_label` (L72)** — extend beyond "Established" (≥3) to the full RAISE scale (e.g. 4.0 "Strong", 5.0 "Optimized"). Latent: today's max is 2.85, so no target even reaches "Established".
2. **`_y_pct` y-axis cap (L124)** — `min(w, 3.0)` clips at 3.0. Currently never fires (max 2.85), and the 0–3 zoom is *deliberate* — the whole roster lives in 0.75–2.85, so a 0–5 axis would squish every dot into the bottom ~57%. If a future target scores >3.0, either lift the cap or make the axis-max data-driven (`max(3.0, ceil(max_w))`) rather than a blind 0–5.
3. **`DEFAULT_BASELINE` (L31) + `rows_table` (L102)** — read the active baseline from the `CURRENT` file (as `test_render.py` does) with a fallback, and thread the resolved `baseline_dir` through `rows_table` for the report-link column instead of the module global, so `--baseline-dir` is honored consistently.

Context: reporting/visualization script only — no finding, score, or frozen baseline is touched.

virtualsteve-exa:
**Partial — baseline-dir consistency delivered in 1.2; full 0–5 RAISE scale carried to 1.3.**

Delivered in 1.2 (on `dev`): the **baseline-dir consistency** half. `suite_health.py` was hardcoded to `v1.1-claude48` (unlike the coverage generators) and so pointed the master public index at the *old* baseline; it now reads the `CURRENT` marker like `owasp_coverage.py`/`raise_coverage.py`, so the page tracks the frozen baseline (now `v1.2-opus5`) and won't drift on re-baseline. It also gained a per-target *remit ↗* link.

Carried to **1.3**: the **full RAISE 0–5 scale** rendering on the suite-health page (the issue's primary ask). Keeping this open for that.

virtualsteve-exa:
[...]

PR #219, opened 2026-08-01
Release 1.2.0 — OWASP 2026, schema 3.0, Opus-5 baseline, community marketplace

virtualsteve-exa:
**Do not merge before the August 4 launch window.** Opening now so the full release diff can be reviewed ahead of the day.

Promotes `dev` → `main` per the governance model: **merge commit, never squash**, then fast-forward `dev` back up so `main` stays an ancestor of `dev`. Tag `v1.2.0` is cut from `main` after the merge; `release.yml` is tag-triggered.

**80 commits · 264 files · +143,648 / −5,714** (the bulk is the re-frozen `v1.2-opus5` baseline set and regenerated renders). The 80th commit is `24dbd42`, syncing `main` back into `dev` after the DCO privacy-alias fix (#221) landed on `main` directly — `dco.yml` is now identical on both branches, so this promotion has nothing to reconcile there.

## What's in it

**Knowledge bases (the headline)** — OWASP **LLM Top 10 2026** + restructured **Agentic AI Top 10 2026** (both from the [OWASP GenAI Security Project](https://genai.owasp.org/)), full renumber sweep. New 2026 categories fire on real targets (LLM08 *Hidden Context Exposure*, LLM09 *Vector & Embedding Weaknesses*), closing the LLM08 under-tag class (#169). Rendering is **version-gated**: a document renders with the vocabulary its findings were tagged under, so a 2025-tagged report is never shown with 2026 names.

**Schema 3.0** — `policy_rule_ids`/`policy_rule_text` parallel arrays; presence now enforced (null allowed, absent rejected).

**Harness reliability** — the watchdog/timeout work (#33/#29) plus the compound-contributor decomposition decidability rule; the Stage-1 gate collapsed uAgents' finding-count spread 4→1.

**Baseline** — fresh **`v1.2-opus5`** freeze, 12 targets, median-of-3, mean median RAISE **1.67**. A post-freeze FP sweep (independent cross-model reviewer per target) found **zero detector false positives** across ~130 findings; remit over-reach fixed and re-frozen on 6 targets, the rest tracked (#198/#200/#201).

**Distribution — community marketplace** — installs move to [`open-agent-ai-security/plugins`](https://github.com/open-agent-ai-security/plugins) for **both Claude Code and Codex** (Codex path smoke-verified on CLI 0.146.0). The in-repo `.claude-plugin/marketplace.json` becomes a CI-guarded legacy mirror so existing installs keep updating; `marketplace-sync.yml` + `check_marketplace_mirror.py` fail closed on drift and pin the canonical praxen source.

**Renderer** — pretty-remit HTML renderer + freshness gate; TXT summary no longer deletes route placeholders like `<id>` (#215 — found by an end-to-end acceptance run, present in 1.1.0 too).

**Contributor** — external hardening contribution from @cdeck-exabeam merged with authorship preserved (#194).

## Release gates (run on this head)

| Gate | Result |
|---|---|
| `./build.sh` (4-manifest version guard + docs freshness + remit gate) | ✅ |
| `tests/render/test_render.py` | ✅ **245 / 0** |
| `check_marketplace_mirror.py` vs live canonical index | ✅ in sync |
| Version consistency — PRAXEN_SPEC / plugin.json / marketplace / codex | ✅ all `1.2.0` |
| Ancestry — `main` is an ancestor of `dev` | ✅ |

## Launch-day sequence

> **Merging is what ships.** The community catalog pins praxen's `main`, so users get 1.2.0 the moment this merges — the tag only cuts the zip and the GitHub release page. Sequence accordingly.

1. **Publish the blog post first** — merge blog PR #21 (install commands), flip `published: yes`, regenerate, push. The news-page card links to it, and after step 2 users can already install.
2. **Merge this PR with a merge commit** (not squash), then `git push origin origin/main:refs/heads/dev` to fast-forward `dev`.
3. **Smoke the install path immediately** — canonical catalog (Claude Code + Codex) and the legacy mirror. This is the first moment real users are served 1.2.0, so verify here rather than after the tag.
4. Tag `v1.2.0` from `main` → `release.yml` cuts the zip and GitHub release; confirm the release artifact.

## Known / deliberate

- **#193** (setup-python 6→7) is parked until after the release — it touches `release.yml`, which no PR run exercises; first execution would be this tag push.
- Deferred to 1.3, tracked in `plans/RELEASE_1.3_PLAN.md`: #48 scoring rework, #198/#200/#201 remit cleanup + re-baseline, #195/#196/#197, #176, #216/#217.


virtualsteve-exa:
# Final pre-release review — praxen 1.2.0 (`dev` → `main`)

## ✅ SHIP

79 commits, 265 files. All five gates green, version chain consistent, CHANGELOG matches the diff, baseline numbers reconcile exactly. **No must-fix-before-tag items.** Nine low/informational items below, all deferrable to 1.2.1.

---

## Gates run (locally, on `origin/dev` @ `8b7a824`)

| Gate | Result |
|---|---|
| `./build.sh` | **exit 0** — `markdown` present so the `guide/` rebuild + freshness check actually ran (not skipped); `render_all_remits.py --check`: *all 15 pretty-remit HTMLs in sync*; → `dist/praxen-1.2.0.zip` (7.8M, 91 files) |
| `python3 tests/render/test_render.py` | **245 passed, 0 failed** |
| `python3 scripts/release/check_marketplace_mirror.py` | **exit 0** — *"marketplace mirror is in sync with the canonical index"* (live fetch of `open-agent-ai-security/plugins@main`) |
| `python3 tests/render/test_plugin_manifests.py` | **18 passed, 0 failed** |
| `python3 tests/render/test_manifest_to_findings.py` | **42 passed, 0 failed** (extra — release.yml runs it) |
| `python3 tests/render/test_scan_diff.py` | **12 passed, 0 failed** (extra) |

Working tree was clean before and after; `git status --porcelain` empty post-`build.sh`. Nothing restored, nothing modified.

---

## 1. Release mechanics — can this ship

**Version chain: consistent.** `1.2.0` in `PRAXEN_SPEC.md:8`, `.claude-plugin/plugin.json:3`, `.claude-plugin/marketplace.json` (both the praxen entry and `metadata.version`), `.codex-plugin/plugin.json:3`, README badge. `build.sh:29-56` guards all four by name (the praxen entry is looked up by name, not position — correct); `test_plugin_manifests.py` re-asserts agreement independently.

**`release.yml` on a `v1.2.0` tag: correct.** Read it line by line, since it has never run against this tree. The spec-version/tag check (`grep -m1 '^\*\*Version:\*\*'`) matches `build.sh`'s own extraction exactly, so they can't disagree. It installs `requirements-dev.txt` *before* `build.sh`, which is what makes the `guide/` freshness backstop actually enforce at tag time rather than silently skip. `gh release create` uploads `dist/praxen-${version}.zip` — the exact path `build.sh` writes. No problems.

**Zip contents: clean.** 91 files / 7.8 MB. No `local/`, no `tests/`, no `plans/`, no `.DS_Store`, no `__pycache__`, no `.git`, no credentials. `graphics/` (7.5 MB) dominates, as in every prior release.

**CHANGELOG `[1.2.0]` vs the actual diff: accurate.** I checked each claim against code, not prose. Delivered and claimed: schema 3.0 parallel arrays (#7), validator rule cross-check (#5), N/A categories, version-gated OWASP rendering, Step 8.5 themes outline (#29), interleaved emission (#33), Step-4 evidence checkpoint + `--validate-manifest` (#65), `SCAN_INSTRUCTIONS.md`, `tests/scan_diff.py`, `render_remit.py`, `marketplace-sync.yml` + `check_marketplace_mirror.py`, `CURRENT` → `v1.2-opus5`, mean median RAISE 1.67. **Nothing claimed-but-not-delivered; nothing delivered-but-unclaimed.**

---

## 2. Correctness of the changed code — no blocking defects

I went after the three areas you flagged plus the surrounding blast radius.

**`_llm_titles_for` (`render.py:113-124`) — correct.** The tuple/conditional precedence parses as intended (`major, minor = int(parts[0]), (int(parts[1]) if len(parts) > 1 else 0)`), and any garbage version degrades to `(0, 0)` → 2025 rather than raising. `_OWASP_LLM_TITLES_2026` matches the `knowledge/KB_LLM_TOP10.md` `## LLM01`–`## LLM10` headings **verbatim**, including "Data and Model Poisoning" and "Vector and Embedding Weaknesses" — I diffed the two lists rather than eyeballing them.

Both branches are locked by byte-compared goldens, which is stronger coverage than a unit test would be: `tests/fixtures/finbot.canonical.json` is schema 3.0 / `praxen_version` **0.3.0**, so `finbot.golden.html` pins `OWASP LLM Top 10 (2025)` and `finbot.golden.txt:223` pins `Built on OWASP Top 10 for LLM Applications 2025`; the 12 baselines are `1.2.0` and pin the 2026 side. A regression in the gate breaks the render suite.

`_canon_owasp_tag_label` was correctly repointed to `_OWASP_LLM_CODES` (`render.py:947`) — `grep '_OWASP_LLM_TITLES\b'` returns zero dangling references, so no `NameError` waiting in a cold path.

**`strip_tags` (`render.py:238-260`) — the narrowing is safe; I verified nothing relied on the old broad strip.** `_STRIPPABLE_TAGS` resolves to `['code', 'em', 'p', 'strong']`, which is the *exact* union of every allow-list in `_RICH_FIELDS`. There are only three call sites — `behavior_summary` (`render.py:762`), `finding['summary']` (`:821`), `recommended_actions` (`:826`) — and all three fields draw their markup from `_RICH_FIELDS`, so every tag those fields may legitimately carry is still stripped. Confirmed empirically: `'POST /api/vendors/<id>/invoices'` now survives intact, and the fix is pinned in `tests/fixtures/finbot.golden.txt` (`10/min on /vendors//invoices` → `10/min on /vendors/<id>/invoices`).

Residual behavior change, which I checked rather than assumed: tags carrying attributes (`<code class="x">`) and non-allowlisted tags (`<a href>`, `<script>`) now appear **literally** in the TXT instead of vanishing. That is consistent, not a leak — `render_rich` (`:203-232`) protects only bare `</?tag\s*>` and `html.escape`s everything else, so the HTML report already shows those as visible text. TXT and HTML now agree. No XSS surface (TXT is plain text; HTML still escapes).

**Schema 3.0 required-field enforcement (`schema.py:284-309`) — sound.** `_err` (`:106-107`) **raises immediately** rather than accumulating, which resolves the sequencing hazard I went looking for: a manifest with `policy_rule_ids` present but `policy_rule_text` absent aborts on the missing-key check and can never reach `_str_list(f["policy_rule_text"], …)` and a raw `KeyError`. That is precisely the #205 class, and it is closed on both ends — `manifest_to_findings.py:980-984` also now sets *both* keys explicitly to `None`.

Renormalization (`schema.py:494-500`): the `/ wsum` divisor cannot be zero — `NA_ELIGIBLE_KEYS` excludes the two presence-scored categories, and `_validate_raise_posture` enforces all six keys present exactly once with fixed weights before consistency runs.

The `scan_timestamp` tightening (`schema.py:220-231`) is a genuine constraint change, so I traced the producer: `SKILL.md:141` emits `date -u +%Y-%m-%dT%H:%M:%SZ`, which satisfies `_TIMESTAMP_RE`. No self-inflicted breakage.

**`manifest_to_findings.py`** — adding `"n/a"` to `_is_null_sentinel` only reaches `*_or_none` fields (`score`, `line`, `finding_id`, `description`, `policy_rule_*`, `owasp_llm`, `owasp_agentic`). There is no field where a literal `"N/A"` is meaningful content, so the widening is safe.

*Minor robustness nit, not a defect:* `validate_manifest` (`:1132-1197`) catches only `ManifestError` around the section dispatch; a section parser raising e.g. `IndexError` on adversarial input would traceback instead of reporting cleanly. Cosmetic — it's a developer tool on operator-authored input.

---

## 3. Distribution / supply chain — coherent

Fetched the live canonical index. Its praxen entry pins `{"source":"url","url":"https://github.com/open-agent-ai-security/praxen.git","ref":"main"}`, byte-matching `EXPECTED_CANONICAL_PRAXEN_SOURCE` in `check_marketplace_mirror.py:44-48`. The mirror checker is well built — it compares by **full equality minus documented exemptions** rather than an allow-list, so a field the canonical index grows later surfaces as drift instead of slipping through, and the praxen-`source` exemption is correctly asymmetric (the canonical pin must match exactly; only the mirror's `./` is ignored). `marketplace-sync.yml` fires on `.claude-plugin/**` PRs, pushes to `dev`/`main`, and weekly — the weekly run is what catches canonical-side edits no PR here would trigger.

What a user actually gets after merge: coherent. Both paths (canonical `plugins` index, legacy per-repo mirror) resolve to this repo's `main`, marketplace name and plugin key are unchanged, so existing installs are unaffected — as the CHANGELOG claims.

---

## 4 & 5. Docs and baseline

**Docs/site:** every copy-pasteable install command across README, `index.html` (both the terminal block and the copy-button string), `docs/installation.md`, `guide/installation.html`, `docs/index.md`, `CONTRIBUTING.md`, `tests/README.md` and `plugin-smoke.sh` points at the canonical index and they agree with each other. Legacy-mirror mentions are all correctly framed as legacy. No live surface still says "OWASP LLM Top 10 2025" — the only 2025 strings are the intentional "2025 → 2026: what changed" mapping table and a publication date. `guide/` is in sync with `docs/`. I confirmed the anchor the report template now links to (`owasp.html#owasp-top-10-for-llm-applications-2026`) actually exists in `guide/owasp.html`.

**Baseline:** `tests/baselines/CURRENT` = `v1.2-opus5`. All 12 targets present with all four artifacts each; all 12 at `schema_version "3.0"` / `praxen_version "1.2.0"`. Rather than trust `BASELINE.md`, I recomputed Σ(score×weight)/Σ(weight) from every frozen JSON: **all 12 match their declared `weighted_overall` to 4 dp**, and all 12 match `BASELINE.md`'s table exactly. Mean = 20.05/12 = **1.6708 → 1.67**, matching both `BASELINE.md:46` and `CHANGELOG.md:32`. Spot-checked uagents (1.70, 12 findings), salesforce (1.70, 12), openhands (2.00, 14) by hand — all exact. `examples/README.md`'s refreshed counts match the example JSONs exactly (finbot 16/6C-3H-7M/0.90; helperbot 13/3C-4H-5M-1L/0.60; salesforce 12/6H-4M-2L/1.70). No secrets, no stray files in the frozen set.

*Worth knowing, not a defect:* **no category is N/A anywhere in this baseline**, so the schema-3.0 renormalization path ships exercised only by the synthetic tests in `test_render.py` / `test_manifest_to_findings.py`. Those tests are good (null accepted on vector-scored, rejected on presence-scored, non-renormalized overall rejected) — but the first real N/A will be seen in production, not in the freeze.

---

## Findings

### Must-fix before tag
**None.**

### Can follow in 1.2.1

**L1 — `tests/baselines/README.md:8` names the wrong predecessor set.** Says the retained archives are `v1.0.2-claude48/` and `v0.7.7-claude48/`, omitting `v1.1-claude48/` — the set v1.2 actually supersedes (`BASELINE.md:8` says so). Line 35 of the same file gets it right. *Scenario:* a contributor doing a regression comparison picks `v1.0.2` as the baseline-to-beat and compares against a two-releases-stale set.

**L2 — FP-sweep denominator is off, or ambiguous.** `CHANGELOG.md:32` and `tests/baselines/v1.2-opus5/BASELINE.md:64` both say the sweep covered "~130 findings"; the frozen set contains **156**. Either the sweep ran against pre-fix reports (true but ambiguously worded) or the claim understates coverage by 26. *Scenario:* someone cites "zero FPs across ~130 findings" externally and the number doesn't reconcile against the published baseline.

**L3 — dead docs anchor for every report generated by 1.1.x.** `docs/owasp.md` no longer has a heading yielding `owasp-top-10-for-llm-applications-2025` (confirmed: zero hits in `docs/` and `guide/`), but all previously-published reports — including the in-repo `tests/baselines/v1.1-claude48/*/*.html` — link tag chips to `owasp.html#owasp-top-10-for-llm-applications-2025`. *Scenario:* a user opens a report they generated last month, clicks an LLM tag chip, and lands at the top of the OWASP page instead of the LLM section. Fixable with a one-line legacy anchor in `docs/owasp.md`.

**L4 — stale version floor in the install confirm comment.** `docs/installation.md:26` and `:45` (mirrored `guide/installation.html:292,307`) say `# confirm: praxen@open-agent-ai-security, enabled, v1.0.0+`. Cosmetic; reads as pre-1.2 boilerplate on the page a new user follows first.

**L5 — marketplace-removal parenthetical is now wrong for the canonical path.** `docs/installation.md:118` / `guide/installation.html:351`: "removed by its registered name (`open-agent-ai-security`, from `.claude-plugin/marketplace.json`) — which here matches the repo owner used to add it." For a user who followed the canonical path the name comes from the `plugins` repo's manifest, and the repo they added is `plugins`, not the owner. *Scenario:* a user trying to uninstall reasons from the parenthetical and guesses the wrong argument.

**L6 — the shipped zip bakes an absolute home directory into all three examples.** `examples/*/*-findings.json` `scan.workspace` = `/Users/steve.wilson/Documents/github/deckard/local/v1.2-owasp2026-baseline/src/…`. **Pre-existing** — `origin/main` has the same shape with different paths — so not a regression and explicitly not a ship blocker. Worth redacting to a placeholder at some point; it's in the release artifact every user downloads.

**L7 — `release.yml` runs 3 of the 5 suites.** It omits `test_scan_diff.py` and `check_marketplace_mirror.py`. Low impact: CI covers `scan_diff` on every PR and `marketplace-sync.yml` runs on push to `main`, so the gap only bites on a tag cut from an unusual state. Cheap to close by adding both to the tag job.

**L8 (informational) — spec now under-describes the renderer.** `PRAXEN_SPEC.md:398` documents the LLM coverage heading as a static "(2026)", but `report_template.html:1158` is version-gated `{{OWASP_LLM_YEAR}}`. (The adjacent hardcoded Agentic "(2026)" is *correct* — there has only ever been a 2026 Agentic list, and `_OWASP_AGENTIC_TITLES` is deliberately ungated.)

**L9 (informational, no action for this tag) — the merge *is* the release for marketplace users.** The canonical index pins `ref: main`, so every marketplace install tracks `main@HEAD`; the `v1.2.0` tag only produces the GitHub Release zip. Two consequences worth holding in mind: (a) there is no window between merge and tag in which users are still on 1.1.0 — the promotion ships instantly; (b) a marketplace install **clones the repo**, which this release grows from **28.0 MB → 35.6 MB** tracked (`tests/` is 25.6 MB, 72% of the tree). `build.sh`'s careful `INCLUDE` curation — 91 files, 7.8 MB — governs only the zip, which is *not* the primary install path. Not a 1.2.0 problem; a trajectory worth watching before the frozen sets grow again in 1.3.

---

*Recommendation: tag `v1.2.0` as planned. The gates are real and they pass, the numbers reconcile independently, and the three riskiest code changes (schema-3.0 required fields, the OWASP version gate, the `strip_tags` narrowing) are each pinned by byte-compared goldens rather than by assertion alone.*

*Not approving or merging — that's Steve's call.*

— final pre-release review agent, requested by Steve

virtualsteve-exa:
## ✅ GO — cleared for the Aug 4 merge. Zero must-fix items.

Final go/no-go pass over `origin/main...origin/dev` @ `24dbd42`. Scope was deliberately narrow: (A) re-verify no regression since the prior SHIP verdict, (B) adversarially attack the new DCO change, (C) release-gate spot checks, (D) launch-sequence ordering. I did **not** re-derive the baseline audit. Read-only throughout — no tracked file was modified (`git status` clean at start and finish; `build.sh` only wrote `dist/`, which is gitignored at `.gitignore:6`).

---

## A. Regression check on the `main`→`dev` sync — clean

The sync merge `24dbd42` is exactly what it claims to be and nothing more.

| Check | Result |
|---|---|
| Files changed by `24dbd42` vs first parent | **`.github/workflows/dco.yml` only** (+26/−2). Nothing else rode in. |
| `dco.yml` identical on both branches | ✅ blob `63cd38d` on `origin/main` **and** `origin/dev` |
| `main` is an ancestor of `dev` | ✅ (`branch-drift.yml` invariant holds) |
| Commits on `main` not in `dev` | none |
| PR CI | ✅ all green — `dco`, `check`, `docs-freshness`, `test (py3.9/3.12/3.13)` |

**The PR body's headline stats are now stale** — it says *79 commits · 265 files · +143,653 / −5,717*; actual is now **80 commits · 264 files · +143,648 / −5,714**. The delta reconciles exactly and benignly: `+1` commit is the sync merge itself, and `−1 file / −5 / −3` is `dco.yml` **dropping out of the release diff** because `dev` and `main` now carry identical copies. Pre-sync (`69c0b88...8b7a824`) measured 265 / +143,653 / −5,717 — a byte-for-byte match with the PR body, confirming nothing else moved. Cosmetic only, but the PR body is the release record — worth a one-line edit before merge.

## B. Adversarial review of the DCO change (#221) — **safe to ship**

I extracted the workflow's JS verbatim and ran **17 cases**, including every attack in scope. Bottom line up front:

> **The name-only path grants no forgery capability that the pre-existing exact-match path didn't already grant.** DCO here is a *self-consistency check* over git ident fields that are 100% contributor-controlled — it is not, and has never been, identity authentication.

The decisive comparison:

| Case | Result |
|---|---|
| ATTACK — impersonate "Steve Wilson" using attacker's own noreply alias (**new path**) | `PASS(name-only)` |
| BASELINE — same impersonation via `user.email=steve.wilson@exabeam.com` + exact trailer (**pre-existing path**) | `PASS(exact)` |

The baseline attack is *strictly worse*: setting the author email to Steve's real address makes GitHub attribute the commit to **Steve's actual account** in the UI. The new path requires an `@users.noreply.github.com` alias, which embeds the attacker's own numeric GitHub account ID — **more forensic trace, not less.**

All adversarial negatives still fail correctly:

```
FAIL  no trailer at all
FAIL  trailer names someone else
FAIL  normal author, mismatched email
FAIL  non-noreply author, name-only would-be match
FAIL  lookalike domain  bob@evil-users.noreply.github.com.attacker.io
FAIL  git -c user.name="x[bot]"   <- #209 hardening holds
```

Specific vectors you asked about, resolved:

- **Homoglyph / unicode in the name comparison** (`dco.yml:109`) — fails **CLOSED**. Cyrillic `е` ≠ Latin `e`, so a homoglyph can only reject a legitimate contributor, never admit an illegitimate one. Not a bypass.
- **`c.commit.author.name` containing `<` or a newline to break the trailer regex** (`dco.yml:113`) — **not reachable.** Verified empirically: git sanitizes idents, stripping `<`, `>`, and newlines. Committing with name `Evil <hack@x.com>\nSigned-off-by: Fake` normalizes to `Evil hack@x.comSigned-off-by: Fake`. No trailer injection via the name field.
- **noreply regex** (`dco.yml:108`) — dots properly escaped, end-anchored. `a@b.users.noreply.github.com` and the lookalike domain above both correctly fail.
- **Multiple trailers, one matching** — passes, correct and standard.
- **Case / whitespace via `trim().toLowerCase()`** — deliberate, symmetric, matches conventional DCO tooling.
- **250-commit cap fails CLOSED** (`dco.yml:70-79`) — directly load-bearing for an 80-commit release PR. Good control.
- **`pull_request_target` usage** — never checks out PR code; reads API metadata and comments only. Correct, safe pattern.
- **#221 landed via PR**, not a direct push. `main` protection confirmed: 1 required review, required checks (`dco`, all three `test` jobs), `strict: true`, no force-push, no deletions, `required_linear_history: false` (so the merge-commit promotion model works).

---

## Findings — **nothing blocks the merge**

### Must-fix before merge

**None.** Every release gate passes and the DCO change is sound.

### Post-launch (ranked)

**1. MEDIUM — the bot exemption trusts email-derived attribution rather than an allowlist.** `.github/workflows/dco.yml:87-88`

```js
const ghIsBot = c.author && c.author.type === 'Bot';
if (ghIsBot) continue;                                    // bot account — exempt
```

`c.author` is GitHub's resolution of the **commit author email** to an account. *Failure scenario:* a contributor sets `user.email` to a bot's published noreply address (e.g. `49699333+dependabot[bot]@users.noreply.github.com`); if GitHub attributes that commit to `dependabot[bot]`, `type === 'Bot'` is true and the commit **skips DCO entirely with no sign-off at all** — a strictly better outcome for an attacker than any trailer forgery, since it bypasses the check rather than satisfying it. My harness confirms the code path (`EXEMPT-bot`); I could not safely confirm GitHub's attribution behaviour for bot noreply addresses from here, so treat this as *unverified but cheap to close*.

To be explicit: **this is pre-existing and #221 made it strictly better** — it deleted the `nameIsBot` regex that was the softer half of this exemption. Suggested hardening: gate on an allowlist of `c.author.login` (`dependabot[bot]`, …) instead of accepting any `type === 'Bot'`.

**2. LOW — merge commits are wholly exempt.** `dco.yml:83`

`if (c.parents.length > 1) continue;` For the merge-commit promotion model this is *necessary* and correct — the 1.2 promotion merge will itself be exempt, as will `24dbd42`. The edge worth knowing: an "evil merge" (conflict resolution or hand-edits made in the merge commit) introduces content that appears in no parent and is therefore certified by nobody. This matches standard DCO-bot behaviour and I would not change it for 1.2 — just naming it since the promotion model leans on it.

**3. LOW — `Signed-off-by:` matches anywhere in the message body.** `dco.yml:92-94` (exact path), `dco.yml:112-115` (name path)

Matching is line-based over the whole message, so a trailer inside a fenced code block or quoted text satisfies the check — confirmed on **both** paths, so pre-existing, not new. *Failure scenario:* a docs commit that merely *illustrates* a sign-off passes DCO without the author intending to certify. Low real-world impact (the author still typed the text).

**4. LOW — commit subject interpolated unescaped into the bot comment.** `dco.yml:120` and `dco.yml:143`

```js
const list = bad.map(b => `- \`${b.sha}\` — ${b.subject}`).join('\n');
```

`b.subject` is attacker-controlled. *Failure scenario:* a crafted subject injects markdown — most usefully a phishing link — into a comment posted by `github-actions[bot]`, lending attacker-supplied text bot authority on a security project. Newline injection is impossible (`split('\n')[0]`), so a wholly fabricated "✅ passed" block is out of reach. Wrap the subject in backticks or escape it.

**5. NIT — PR body stats stale.** See section A. Also worth adding one line stating that *merging* is what ships to users (see D) — the body currently frames the tag as the release event.

---

## C. Release gates — all green (run locally on this head)

| Gate | Output |
|---|---|
| `./build.sh` | **exit 0**; 4-manifest version guard + docs freshness + remit gate all pass |
| `python3 tests/render/test_render.py` | **245 passed, 0 failed** |
| `python3 tests/render/test_manifest_to_findings.py` | **42 passed, 0 failed** |
| `python3 tests/render/test_plugin_manifests.py` | **18 passed, 0 failed** |
| `python3 scripts/release/check_marketplace_mirror.py` (**live** canonical index) | `marketplace mirror is in sync with the canonical index.` exit 0 |
| `tests/baselines/CURRENT` | `v1.2-opus5` ✅ · 12 target dirs present ✅ |
| Version consistency | `PRAXEN_SPEC.md:8` · `.claude-plugin/plugin.json` · `.claude-plugin/marketplace.json` (both keys) · `.codex-plugin/plugin.json` — **all `1.2.0`**; README badge agreement covered by `test_plugin_manifests.py` |
| CHANGELOG `[1.2.0]` date | **2026-08-04** ✅ (matches the blog draft's `article:published_time`) |

I deliberately ran **all three** test files `release.yml` invokes, not just `test_render.py` — the tag push is that workflow's first-ever execution (#193 parked), so a failure in either of the other two would fail the release job after the merge has already shipped to users. Both pass. Local `markdown` is **3.9**, matching the `requirements-dev.txt` pin, so the docs-freshness result is representative of CI.

**Zip payload audited** — 91 files. No `local/`, `tests/`, `plans/`, or `.git`; no `.env`/secret/credential/key/token-shaped entries. Top level is exactly `.claude-plugin/ .codex-plugin/ docs/ examples/ graphics/ skills/ CHANGELOG.md CONTRIBUTING.md LICENSE NOTICE PRAXEN_SPEC.md README.md WORKER_REMIT_template.md`.

**CHANGELOG claims spot-checked against the diff, no contradictions:** version-gated rendering (`render.py:130`, `>= (1,2)` selects the 2026 titles), #215 route-placeholder survival (`render.py:249-251`), `tests/scan_diff.py`, `render_remit.py`, `marketplace-sync.yml` + `check_marketplace_mirror.py` all present.

## D. Launch-day sequence — one ordering hazard, easily fixed

**Verified, not assumed:** the live canonical index pins praxen as `{"source":"url","url":"https://github.com/open-agent-ai-security/praxen.git","ref":"main"}`. The legacy in-repo mirror uses `"source": "./"`, which likewise resolves against `main`. **So step 2 — the merge — is what ships 1.2.0 to users. The tag only cuts the zip and GitHub release.**

- **Step 1 (blog first) is correct, and load-bearing.** `news.html:59` and `news.html:223` on `dev` hard-link `https://open-agent-ai-security.github.io/blog/praxen-1-2-release/`, and merging publishes that news page. The post is currently a live **draft** — it returns 200 but carries a "Draft — not yet published" banner, is `noindex`, and is unlinked from the index/feeds/sitemap. Merge before flipping it and the shipped news card points at a page stamped *Draft*. Ordering is right. (Minor: the draft is already publicly fetchable — unlisted, not secret. Deliberate, just be aware the 1.2 narrative is reachable now.)

- **⚠️ Step 4 is the hazard.** The plugin smoke test is labelled *post-tag*, but users can install 1.2.0 the instant step 2 completes — **before** any install has been exercised. As written, the smoke test is post-hoc verification, not a gate, and the unverified-exposure window spans merge → tag → smoke. Compounding it: the canonical-catalog journey **doesn't need the tag at all**, since propagation on a `ref: main` pin is merge-driven. The older "pre-tag still shows the previous version" habit came from the per-repo/tag install model and **no longer applies to either catalog path**.

  **Recommendation:** split step 4 — run the canonical-catalog and Codex install smoke **immediately after step 2, before step 3**, and keep only the zip / GitHub-release checks post-tag. That shrinks the window to merge → smoke.

- **Keep the step 2 → step 3 gap short.** Between them, users installing get 1.2.0 while the GitHub "Latest release" still reads 1.1.0 with no v1.2.0 zip — so anyone arriving from the blog's *Get Praxen 1.2* section hits a stale releases page.

- `release.yml` confirmed on `setup-python@v6` / `checkout@v7` with the tag↔`PRAXEN_SPEC.md` version guard intact — #193 correctly parked.

---

**Is the DCO change safe? Yes.** It is a narrow, well-reasoned relaxation that grants no new forgery capability, fails closed on every adversarial case I could construct, and on balance *improves* the workflow by removing the contributor-controlled `nameIsBot` exemption. The residual weakness is the pre-existing bot-attribution exemption (finding 1), which #221 did not introduce and partially mitigated.

No approval and no merge performed, per the review gate.

— final go/no-go reviewer, requested by Steve

Issue #224, opened 2026-08-03
SKILL.md Step 9.9: --validate-manifest example is unrunnable, and skeleton-tolerance wording overstates the case


virtualsteve-exa:
**Severity:** Low (documentation) · **Found by:** post-release live sanity check of 1.2.0 (three scans from a marketplace install; both scan agents hit these and worked around them without incident)

Two issues in the same sentence, `skills/behavior-verifier/SKILL.md:800`:

## 1. The documented command cannot run

SKILL.md shows:

```
manifest_to_findings.py --validate-manifest ./reports/<slug>-draft-<TIMESTAMP>.md
```

But `manifest_to_findings.py` (lines 1261, 1266) declares:

```python
ap.add_argument("--manifest", required=True, metavar="PATH", ...)
ap.add_argument("--validate-manifest", action="store_true", ...)
```

`--validate-manifest` is a **flag**, not an option that takes a value, and `--manifest` is **required**. The documented form fails twice over: the path is parsed as an unrecognized positional, and the required `--manifest` is missing.

Correct form:

```
manifest_to_findings.py --manifest ./reports/<slug>-draft-<TIMESTAMP>.md --validate-manifest
```

**Failure scenario:** an executing model follows Step 9.9 literally mid-composition, gets an argparse usage error, and either burns turns debugging it or skips the mid-flight validation the step exists to encourage — losing the early format check *and* the watchdog reset it doubles as.

## 2. "designed to pass on an incomplete skeleton" overstates the tolerance

The tolerance is real but narrower than the wording implies. `validate_manifest` separates `errors` from `missing` sections and only runs the full pipeline + schema checks once **nothing is missing** — so a skeleton with sections *absent* passes, as documented.

But a section that is **present and empty** goes down the full-validation path, where `findings.schema.json` sets `minItems: 1` on `remit_coverage.rules` (and on `findings[].tags`, `policy_rule_ids`, `policy_rule_text`, `evidence`, `recommended_actions`). A skeleton that has written the `remit_coverage` header but no rules yet is therefore rejected.

The behavior is arguably correct — it's the wording that misleads. Suggest something like: *"passes while sections are still absent; a section that is present but empty is validated in full."*

## Suggested fix

One-line correction to the command plus a clause on the tolerance. Documentation only — no scan-behavior change, no baseline impact.

*Filed by Claude Fable 5 on behalf of Steve Wilson*

virtualsteve-exa:
[...]

PR #233, opened 2026-08-05
Praxen 1.2.1 (rolling) — held until release-ready

virtualsteve-exa:
First PR of the 1.2.1 fast-follow (plan: \`plans/RELEASE_1.2.1_PLAN.md\`, now ACTIVE). Section **B** in full, plus the plan-status flip.

## What's here

**\`scripts/release/bump_version.py\`** (new) — ports socxen's bump script. Edits all **six** version surfaces (the draft plan said five; \`marketplace.json metadata.version\` surfaced when porting against \`test_plugin_manifests.py\`'s actual check list), verifies each by parsing after the edit, never commits. By-name marketplace lookup with a stray-version precondition; shields \`--\` prerelease escaping; docs/ version-string tripwire. Round-trip tested 1.2.0→1.2.1: exactly six version strings changed, manifest test 18/18 on the bumped state, then reverted.

**#209 item 3 (+ #193 folded in)** — every \`uses:\` across all seven workflows SHA-pinned with a version comment: checkout **v7.0.1** (export-stargazers rides along v4→v7), setup-python **v7.0.0** (the #193 bump — now also covering \`marketplace-sync.yml\`, which #193 predates), github-script v9.0.0, fetch-metadata v3.1.0. Dependabot keeps pins current.

**#209 item 2** — dependabot auto-merge now verifies the base branch has **required status checks** before enabling \`--auto\`, failing closed with an explanatory comment. Without required checks, GitHub's auto-merge merges *immediately* instead of waiting for CI — and that is \`dev\`'s live state today (zero required checks), so this closes a real instant-merge hole.

**#222** — DCO bot exemption now requires the GitHub-resolved \`Bot\` type **and** membership in an explicit login allowlist (\`dependabot[bot]\`).

## Recorded risk

\`release.yml\` is tag-triggered — the 1.2.1 tag push will be setup-python v7's **first execution** there. A failure needs a *new tag* (re-runs replay the old workflow definition); acceptable on a patch, recorded in the plan.

## Effect on Steve's workflow

Auto-merge on \`dev\` is now **off in practice** until dev gets required status checks — Dependabot PRs will carry a comment explaining why and merge manually on green.

Closes #193. Closes #222. Part of #209 (items 2–3).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #227, opened 2026-08-03
Report tag chips link the unstyled Jekyll /docs/ render instead of /guide/ — and .nojekyll would break reports already in the wild


virtualsteve-exa:
**Severity:** Low (degraded experience, not broken) · **Deferred deliberately** — surfaced during the 1.2 promotion and held out of the release; no user-facing breakage justifies touching the renderer mid-promotion.

Split out of the docs review (#226), which proposed adding `.nojekyll`. That fix would be actively harmful — see "Do not add .nojekyll alone" below.

## What

Generated reports link Praxen's framework docs at `https://open-agent-ai-security.github.io/praxen/**docs**/…`. That path is GitHub Pages auto-rendering the raw Markdown through Jekyll: unstyled, off-theme, no site chrome. The intended destination is `/praxen/**guide**/…`, the generated styled pages built by `docs_build.py`.

Both return **200** — verified — so nothing is broken; readers just land on a worse page than the one we build for them.

Source of the bad base (4 occurrences, 2 files):

- `skills/behavior-verifier/render.py:400` — `_DOCS_BASE`, which feeds `_tag_href()` for **every** finding tag chip (`owasp.html#llm02`, `RAISE.html#<slug>`, the MCP anchor)
- `skills/behavior-verifier/report_template.html:1062,1210,1288` — the findings-intro and RAISE section prose

Anchor schemes are compatible: `guide/owasp.html` carries `id="llm01"`…`id="asi0N"`, so repointing needs no anchor rework.

## Blast radius (why this is not a one-line change)

**85 committed HTML artifacts** carry the old base — `tests/baselines/`, `examples/`, `tests/fixtures/`. The byte-render gate requires every committed render to reproduce exactly from its JSON, so changing the source means regenerating all 85.

That is a **re-render, not a re-scan**: no findings JSON is touched, so no finding, severity, score, or baseline value moves. Deterministic, and each diff should be *exactly* the URL substitution — mechanically verifiable.

## Do not add `.nojekyll` alone

The review's suggested fix makes `/praxen/docs/` return **404** instead of merely unstyled. Every report already generated and distributed by users at 1.0/1.1/1.2 links that path — so `.nojekyll` converts a cosmetic issue into broken links in artifacts already in the wild, which we cannot reach or regenerate.

## Suggested fix

1. Repoint `_DOCS_BASE` and the three template links to `/praxen/guide/`
[...]

Issue #225, opened 2026-08-03
Docs review (1.2 landing): schema 2.0 stale in the stability contract + reports reference, v1.2.0 untagged, scan-diff undocumented


virtualsteve-exa:
Scope: did the 1.2 changes land in the docs — repo (`main` @ 784203d) and the live GitHub Pages site. Cross-document consistency (terminology, tone, structure, dead links) is a separate reviewer's job and is deliberately not covered here.

**Live site matches the repo on every page checked** — no deployment lag. `guide/*.html` is also in sync with `docs/*.md` (re-ran `docs_build.py`; zero diff).

Verified clean, so nobody re-checks these:

- **OWASP 2026 landed cleanly.** Every current user-facing surface says 2026 and uses 2026 titles. `docs/owasp.md` LLM01–LLM10 and the ASI table match `KB_LLM_TOP10.md` / `KB_AGENTIC_TOP10.md` exactly. The only surviving "2025" anywhere in the reviewed surfaces is the deliberate `docs/owasp.md:47-62` "2025 → 2026: what changed" mapping table — correct and worth keeping. No 2025-era category titles ("Insecure Plugin Design", "Overreliance", "Model Theft", "System Prompt Leakage" as a live title) survive in prose, headings, alt text, or captions.
- **Install paths are on the community marketplace everywhere.** `README.md:27-28`, `docs/installation.md:24-25,43-44`, `docs/index.md:81`, `CONTRIBUTING.md:189,194`, `index.html:571-572`, and the front page's copy-install **JS** (`index.html:661-662`) all target `open-agent-ai-security/plugins`. The JS matches the visible terminal block. Both migration notes for the old `open-agent-ai-security/praxen` marketplace are present and correct (`docs/installation.md:29-34` Claude, `:50` Codex).
- **Command shapes are valid.** Checked every documented verb against the real CLIs (`claude plugin --help`, `claude plugin marketplace --help`, `codex plugin --help`, `codex plugin marketplace --help`) — nothing installed. `add`/`install`/`list`/`update`/`uninstall`/`remove` for Claude and `add`/`list`/`upgrade`/`remove` for Codex all exist as documented, including the Claude `update` vs Codex `upgrade` asymmetry in `docs/installation.md:67,76`.
[...]

PR #234, opened 2026-08-05
1.2.1 section A: documentation corrections (#224, #225/#226 remainder)

virtualsteve-exa:
Second PR of the 1.2.1 fast-follow (plan section **A**). Docs only — no scan surface except the #224 command correction inside SKILL.md's already-optional Step 9.9 validation aside.

- **tests/README.md** — dropped the stale "schema-2.0" qualifier on the baseline sweep (it validates every committed set; current is 3.0), and re-pointed the release-review compare from the **archival** `v1.1-claude48` to `v1.2-opus5`, with a deferral to `baselines/CURRENT` as the authority so this sentence can't go stale the same way twice.
- **docs/installation.md** — confirm lines `v1.0.0+` → `v1.2.0+`; new **"Upgrading from 1.1?"** callout: schema 3.0 arrays (`policy_rule_ids`/`policy_rule_text`), 2026 OWASP names, scores not comparable to 1.1, legacy-marketplace migration pointer.
- **docs/understanding-variability.md** — `tests/scan_diff.py` documented where the manual theme/rule-text diff guidance already lives: what it joins on (remit-anchored rule refs, not run-local `R-NN`), repo-not-zip location, schema-3.0 requirement.
- **docs/writing-remits.md** — `render_remit.py` invocation now points into `skills/behavior-verifier/` (#225 item 7; the bare-filename form fails from a working directory).
- **SKILL.md Step 9.9 (#224)** — the `--validate-manifest` example is now the runnable form (`--manifest <path> --validate-manifest`), and the skeleton-tolerance sentence matches verified behavior: absent sections pass, but the Step-1 skeleton has every heading present, so the full schema check runs and flags the still-empty `remit_coverage.rules` (`minItems: 1` — the only such constraint; empty `findings[]` is legal). Guidance is now "validate after the first rule lands."
- `guide/` rebuilt from the fresh markdown; `./build.sh` and the render suite green.

Closes #224. With PR #233, this completes sections A and B; what remains of #225/#226 scope rides the later 1.2.1 PRs (#216+#4, then #227/#27/#6).


virtualsteve-exa:
## Automated review (Claude)

Factual-accuracy review of the section-A docs corrections. Every claim in the new text was checked against the code. One material error found; everything else verifies.

### High

- **tests/README.md:17** — the corrected sweep sentence introduces a new wrong fact. It now reads: "each baseline JSON must validate against `schema.py` (which knows every shipped `schema_version`, 3.0 for the current set)". Neither half matches the code:
  - `schema.py` knows exactly **one** version: `SCHEMA_VERSION = "3.0"` (skills/behavior-verifier/schema.py:35) and hard-rejects anything else — "renderer understands schema {SCHEMA_VERSION} exactly" (schema.py:203–205).
  - The sweep does **not** validate every committed baseline: test_render.py legacy-skips any baseline whose `schema_version` != 3.0 from *all* schema/render checks (tests/render/test_render.py:623–624, "frozen legacy artifact, schema/render checks skipped"). The archival sets (`v1.1-claude48` at 2.0, and older) get zero schema validation.
  - Accurate statement: the sweep walks every committed baseline, but only current-schema (3.0) files are schema-validated; non-current sets are exempt legacy artifacts, and a #208 guard (test_render.py:606–612) requires every CURRENT-set file to be at 3.0 so the current set can't silently lose its regression net.

### Low

- **tests/README.md:17, same sentence (retained clauses)** — "each post-relicense one must re-render byte-for-byte … and quote its `tests/remits/<slug>.md` verbatim" is also stale: both the byte re-render gate and the remit-verbatim gate apply **only to the CURRENT set**, not all post-relicense sets (test_render.py ~645: "Only the CURRENT set is byte-gated against the live template"; remit check likewise CURRENT-only). Pre-existing wording, but this PR's stated purpose is fixing this exact bullet — worth folding into the same edit.
- **docs/installation.md callout → interpreting-reports.md link** — the callout tells anyone "parsing the old single-value fields" to see the reports reference, but interpreting-reports.md:161 only *lists* `policy_rule_ids` / `policy_rule_text` in the findings row without saying they are now arrays (contrast `recommended_actions[]`, which is explicitly labeled "(array …)"). Link resolves and the fields are covered, but the array shape isn't called out at the destination. Optional follow-up, not a blocker.

### Nitpicks

- **docs/understanding-variability.md** — the doc says the tool reports "new, resolved, and unchanged"; scan_diff.py's own labels are `NEW` / `DROPPED` (tests/scan_diff.py:159,163), and matched pairs aren't labeled "unchanged" (they can carry severity changes — which the doc itself correctly caveats). Minor terminology drift.
- **"Requires schema-3.0 JSON … on both sides"** — correct advice, but the tool doesn't enforce it: there is no `schema_version` check in scan_diff.py, and a 2.0 file's string-valued `policy_rule_text` would silently iterate as characters instead of erroring. A version guard in the tool would be a nice later hardening; the doc text is fine.

### Verified accurate (checked against code)

- `tests/baselines/CURRENT` contains `v1.2-opus5`; release-review re-point + "CURRENT is the authority" deferral both correct.
- installation.md callout: `schema_version` is `const "3.0"`; `policy_rule_ids` / `policy_rule_text` are `["array","null"]` with `minItems: 1` in findings.schema.json; the legacy-marketplace migration note the callout references exists earlier in the same file. `v1.2.0+` confirm lines correct.
- scan_diff.py CLI matches the doc exactly: two positionals (`RUN_A.json RUN_B.json`), `--threshold` (default 0.40), `--json` (scan_diff.py:180–185); matching is on `policy_rule_text` sets (remit-anchored) plus token-similarity and evidence-file overlap, never `R-NN`. `tests/` is genuinely excluded from the release zip (build.sh:119–126, explicit comment).
- render_remit.py: "Defaults: --out is REMIT.md with a .html suffix" — the writing-remits fix (skill-folder path, output next to the input) is correct.
- **SKILL.md Step 9.9 (#224), the critical one — all four claims verify:** (a) `--manifest <path> --validate-manifest` is the runnable form (`--manifest` is `required=True`, manifest_to_findings.py:1261–1266; the old inline form put the path after a `store_true` flag and cannot run); (b) missing sections alone exit 0 (`_run_validate`, ~1225: "Missing required sections alone exit 0"); (c) the full parse+schema check runs only when there are no structural errors *and* no missing sections (~1237–1241); (d) `remit_coverage.rules` has `minItems: 1` (enforced at schema.py:249) while `findings` may be empty (schema.py:267, "may legitimately be empty"), and the parsers accept empty `### rules` / `## findings` bodies — so a full-headings skeleton passes structure, reaches the schema check, and is flagged on the empty rules array. "Validate after the first rule lands" is exactly right.
- **SKILL.md scope:** the edit is confined to the single Step 9.9 manifest-emission paragraph and only changes the description of an *optional* self-check invocation — no analysis, scoring, or remit-interpretation wording touched. Scan-output risk from this edit is negligible.
- guide/*.html diffs correspond to the three markdown edits, nothing else. One cosmetic note: in guide/installation.html the new callout and the pre-existing "staying current" note now share a single `<blockquote>` (the generator merges adjacent markdown blockquotes) — faithful to the toolchain, just renders as one box.
- DCO: single commit `8a7b3be`, `Signed-off-by: Steve Wilson` present. 8 files changed, all in the PR's stated scope; no unrelated changes.

**Verdict:** one factual error to fix before merge — the tests/README.md schema-sweep sentence claims validation coverage the code doesn't provide; everything else, including the #224 Step 9.9 correction, verifies clean against the code.

virtualsteve-exa:
Review findings addressed in the follow-up commit: the tests/README.md sweep sentence now states the real gate (current-schema sets validate + byte-render; CURRENT additionally remit-verbatim; older sets legacy-skipped — verified at test_render.py:623), interpreting-reports.md calls the rule fields arrays explicitly (parallel-by-index wording verified against the schema description), and the upgrade callout also links STABILITY.md. Build green, suite 245/0.

virtualsteve-exa:
Consolidating: all 1.2.1 work is being held on the rolling `1.2.1` branch (merged there as `ac91afe`) and will go to dev as one unit when the release is further along. This PR's review (comment 5197125549) and its fixes are part of that branch — nothing here is discarded.

PR #236, opened 2026-08-05
mirror: drop socxen from the legacy mirror (#235) — fast-tracked ahead of socxen#66

### LINKED ISSUE #235: Remove socxen from the legacy marketplace mirror — it acquires users onto a deprecated channel
Request from the socxen side (cc socxen#66).

## Ask
Drop the **socxen entry** from praxen's legacy mirror (`.claude-plugin/marketplace.json`) and teach `check_marketplace_mirror.py` that this is a deliberate mirror↔canonical difference (it already normalizes two others). **Keep the praxen entry exactly as is.**

## Why
The mirror exists for **continuity**: praxen has real legacy installs (`marketplace add open-agent-ai-security/praxen`) that must keep working. The praxen entry serves that purpose. The socxen entry does not — socxen has **zero** installs via the legacy channel (its whole install base is its two maintainers, both on the canonical `plugins` index). So the socxen entry's only reachable effect is *acquisition*: a legacy-marketplace user discovering socxen there installs it through a channel slated for retirement and is born needing migration. Dead-end onboarding is worse than no onboarding — the canonical index is the sole path we want new socxen users on.

## Why now
socxen is about to restructure its payload into a `plugin/` subdirectory (socxen#66, `git-subdir` source). If the mirror keeps its socxen entry, that entry must be edited in step (or it breaks and trips your weekly drift cron). Removing it instead permanently decouples praxen's mirror from socxen's layout — praxen never has to track socxen changes again.

## Sequencing
Ideally lands before socxen#66 promotes to `main` (target: this week), but there's no hard ordering — with zero users on the path, a gap is harmless; removal just needs to happen rather than an in-step edit.

virtualsteve-exa:
Fast-tracked to **main** on Steve's call (2026-08-05: separate path, "timing is more urgent" — socxen#66 targets this week and its canonical-index flip would otherwise red praxen's weekly drift cron). Deliberately NOT part of 1.2.1: no version bump, CI/packaging metadata only.

- **Mirror** (`.claude-plugin/marketplace.json`): socxen entry removed; metadata says "carrying praxen only" and points siblings at the canonical index. Rationale per #235 — zero socxen installs on the legacy channel, so the entry's only effect was acquiring users onto a deprecated path, plus permanent coupling to socxen's layout.
- **Drift check** (`scripts/release/check_marketplace_mirror.py`): the omission is now the fourth documented mirror↔canonical difference, asymmetric and fail-closed — mirror may omit non-praxen canonical entries; **praxen is required in the mirror**; mirror-only entries are still drift.

Verified: live check green against the current canonical (which still lists socxen — omission working as intended); negative tests confirm both fail-closed paths still trip (mirror-only entry, deviated canonical praxen source). `test_plugin_manifests.py` 17/17 (the dropped check was the socxen-entry check), `./build.sh` green.

**After merge:** dev needs a merge-back (`main` → `dev` merge commit) to keep the ancestry invariant; I'll do that on request. The plugins-repo side of socxen#66 (validator must learn `git-subdir` before the index flips) is separate and still open.

Closes #235.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -5,7 +5,7 @@
     "url": "https://github.com/open-agent-ai-security"
   },
   "metadata": {
-    "description": "Open Agent AI Security's Claude Code plugin marketplace — legacy mirror. The canonical index is https://github.com/open-agent-ai-security/plugins; this in-repo copy serves existing installs added from open-agent-ai-security/praxen and is kept in sync by CI.",
+    "description": "Open Agent AI Security's Claude Code plugin marketplace — legacy mirror carrying praxen only. The canonical index is https://github.com/open-agent-ai-security/plugins; this in-repo copy serves existing installs added from open-agent-ai-security/praxen and is kept in sync by CI. Other community plugins install from the canonical index.",
     "version": "1.2.0"
   },
   "plugins": [
@@ -20,25 +20,6 @@
         "ai-agent-security",
         "raise-framework",
         "owasp"
-      ],
-      "category": "security"
-    },
-    {
-      "name": "socxen",
-      "source": {
-        "source": "url",
-        "url": "https://github.com/open-agent-ai-security/socxen.git",
-        "ref": "main"
-      },
-      "description": "socxen — agentic SOC analyst. Triages Exabeam New-Scale alerts and cases end to end via the Exabeam MCP: gathers evidence, correlates activity, reaches a threat/false-positive verdict, and acts — containment recommended for human approval, and dismiss/close gated by Claude Code permission rules.",
-      "license": "Apache-2.0",
-      "keywords": [
-        "soc",
-        "soc-analyst",
-        "alert-triage",
-        "exabeam",
-        "ai-agent-security",
-        "mcp"
       ],
       "category": "security"
     }
--- a/build.sh
+++ b/build.sh
@@ -32,9 +32,9 @@ if [[ "$VERSION" != "$PLUGIN_VERSION" ]]; then
   echo "error: version mismatch — PRAXEN_SPEC.md says $VERSION, plugin.json says $PLUGIN_VERSION" >&2
   exit 1
 fi
-# Look the praxen entry up BY NAME: the mirror lists other plugins too (which
-# deliberately carry no version, mirroring the canonical index), so a positional
-# read would crash on a reordering rather than checking what it means to check.
+# Look the praxen entry up BY NAME, never positionally: the mirror is
+# praxen-only today (#235), but a by-name read stays correct if entries are
+# ever added or reordered, rather than silently checking the wrong one.
 MARKET_VERSION="$(python3 -c "import json, sys
 m = json.load(open('.claude-plugin/marketplace.json'))
 e = next((p for p in m['plugins'] if p.get('name') == 'praxen'), None)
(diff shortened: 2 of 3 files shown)


virtualsteve-exa:
## Automated review (Claude)

**Verdict: Approve — fail-closed properties hold under adversarial tracing; findings are cosmetic/pre-existing only.**

### Fail-closed integrity (dimension 1) — verified by execution, not just reading
Built 9 adversarial canonical/mirror fixtures and ran the checker against each:

| Case | Result |
|---|---|
| Canonical **removes praxen entirely** | Caught (exit 1) — praxen becomes a mirror-only entry, trips the `mirror_only` check ("mirror carries entries absent from the canonical index: ['praxen']"). The skipped pin check does not create a silent pass. |
| Canonical renames praxen (praxen2) | Caught — same mirror-only path |
| Mirror missing praxen (empty plugins) | Caught — "mirror has no praxen entry" |
| Mirror carries a rogue entry canonical lacks | Caught |
| Shared entry (praxen) field tampered | Caught — full-equality per-entry compare still runs on the intersection |
| Canonical praxen source hijacked | Caught by the explicit pin |
| Canonical grows a new top-level key | Caught (unchanged behavior) |
| Canonical adds a sibling plugin | Silently permitted — the intended #235 behavior |
| **Mirror praxen source hijacked** (swapped to an external git URL) | **Passes the drift checker (exit 0)** — see Informational below |

Live run against the real canonical (which still lists socxen): exit 0, so the omission path is exercised for real, not just in fixtures.

### Findings

**Low**
1. PR body says `test_plugin_manifests.py` 18/18, but on this branch it is **17/17**. The dropped 18th check is exactly the per-entry required-keys check for socxen emitted by the `for e in plugins` loop (tests/render/test_plugin_manifests.py:94) — nothing else was lost. Body count is just stale.
2. build.sh's comment above the version lookup ("the mirror lists other plugins too") is now stale; the by-name lookup itself is unaffected and passes.

**Informational / pre-existing (not introduced by this PR)**
3. `normalized()` pops the praxen `source` on BOTH sides, so a tampered **mirror** praxen source sails through the drift checker (verified: exit 0). Pre-existing on main (function unchanged), and mitigated elsewhere — test_plugin_manifests.py requires plugins[0].source to be a `./`-relative *string*, so a git-URL dict fails CI. Optional hardening: have the checker pin the mirror source to `"./"` the same way it pins the canonical one.
4. Deliberate trade per #235: the weekly cron no longer acts as a tripwire on canonical-index *additions* (a compromised canonical adding a plugin no longer reds this repo). Documented in the docstring; accepted risk given the mirror's praxen-only purpose.

### Other dimensions — all clean
- **JSON/diff hygiene**: marketplace.json valid; praxen entry identical to origin/main (field-level and canonical-dump comparison); `plugins == ["praxen"]`. No socxen references remain on origin/main outside marketplace.json (removed here), CHANGELOG.md, and plans/.
- **Consumers**: build.sh looks praxen up by name — unaffected. bump_version.py (1.2.1 branch only, confirmed absent from main) is also name-based; its "no other entry carries a version" precondition is now trivially satisfied, and its socxen mentions are provenance comments only. marketplace-sync.yml invokes the checker unchanged.
- **Commit hygiene**: single commit `16dee4a` scoped to exactly the two files; DCO Signed-off-by present and matches the author.

One-line verdict: **Approve** — the asymmetric check is genuinely fail-closed on every dangerous path, including canonical-drops-praxen, which is caught by the mirror-only drift arm.

virtualsteve-exa:
Review's pre-existing gap closed in the follow-up commit: the mirror praxen source is now pinned to `'./'` in the checker itself (negative-tested — a tampered mirror source fails with its own drift line), instead of relying solely on test_plugin_manifests.py. build.sh's by-name-lookup comment freshened for the praxen-only mirror. Also correcting the PR body's test count: the manifest test is 17/17 after the socxen-entry check dropped — the review is right that the dropped check was solely socxen's.

